PR #74, opened 2015-09-03
Add cache manager for ALAsset

SergioEstevao:
Added a manager to resize images and  cache results for ALAssets.

This avoid that resize operations overflow the system and keeps memory footprint low.

How to test:
 * On a real device, preferably with a lot of photos on the camera roll.
 * Open the example project
 * Click on the + button
 * Scroll trough the list
 * Check the memory footprint of the app to see if memory goes up in big spikes.

Needs Reviews: @sendhil

### DIFF
--- a/Pod/Classes/WPALAssetDataSource.m
+++ b/Pod/Classes/WPALAssetDataSource.m
@@ -1,4 +1,5 @@
 #import "WPALAssetDataSource.h"
+#import "WPALAssetImageCacheManager.h"
 
 @interface WPALAssetDataSource  ()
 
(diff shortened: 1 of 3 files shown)


kwonye:
Scrolled through 2,500 photos on iPhone 6 iOS 8.4.1 with only 10.4mb. Looks great on my end.

sendhil:
I'm not sure it's working @SergioEstevao? Here's the graph I got when running the Allocations instrument:
image

SergioEstevao:
@sendhil You should get a lot of allocations because of the thumbnail
generation but the "live memory" being used should be low and constant.

koke:
I'm getting 3 warnings when I build for profiling:

captura de pantalla 2015-09-04 a las 8 10 52

koke:
I tired it on an iPad mini and memory was released after scrolling for a while

captura de pantalla 2015-09-04 a las 8 32 12

sendhil:
@koke @SergioEstevao mine released too if i let it sit there for a while, but what happened if you kept scrolling?

SergioEstevao:
@sendhil  There is NSCache object that releases the thumbnails from time to time. In my tests I scrolled up and down and never saw the memory being used go above 15Mb.

SergioEstevao:
After some more tests and analyser runs I'm pretty confident that this changes don't stress the live memory. So going forward with the merge.

PR #78, opened 2015-10-29
fix #77: Hide the title button when no group is selected

### LINKED ISSUE #77: Title is (null) when user denied access to photo
If the user denies access to the Photos, then an alert is presented. In the background the title of the view is (null) [arrow down].

(null) in title bar

It's a very minor/edge case issue, but I'd say it's a dent to the perfection of the component.
Will be submitting a pull request in a minute.

bagia:
In WPMediaCollectionViewController -refreshTitle, hide the title button if the data source returned a nil selected group.
This typically happens when the user has denied access to the device’s photo library.

This should fix issue #77


SergioEstevao:
Thanks @bagia for the fix.

PR #80, opened 2015-11-04
Fix single selection mode

### LINKED ISSUE #79: Disabling multiple selection doens't work
If multiple selection is disable but the picker callback is not implemented or not dismissing of the picker, you can still select other assets.

SergioEstevao:
Closes #79 

How to test:

Run the sample app
Select the options button on the top left
Disable the multiple selection option
Go back to previous screen and click on the + (Add option)
Select a asset and see if all goes right.

@koke since you found the issue do you mind to have a look on the PR to fix it?


koke:
Works for me :shipit: 
As mentioned in Slack, if you don't implement a delegate, when you tap on a photo then tap on the camera it crashes, but that's probably a different issue

PR #81, opened 2015-11-09
Add accessibility to Media collection view

bagia:
ref [wordpress-mobile/WordPress-iOS#4352](https://github.com/wordpress-mobile/WordPress-iOS/issues/4352) - currently the thumbnails in the media collection view are not accessible. Voice Over won’t be able to get to them because the image views “isAccessibilityElement” property is set to NO. As far as the accessibility label is concerned, the choice made here is to stay simple: just get the index of the asset in the view, and - if available - its type if it’s image or video.

At some point I considered leveraging the caption label for accessibility, but it is actually a deceiving name as it is used only display the duration of video items. Note that since accessibility is now enabled at the UIImageView level in this commit, all the subviews including the caption label will become accessible.

### DIFF
--- a/Pod/Classes/WPMediaCaptureCollectionViewCell.m
+++ b/Pod/Classes/WPMediaCaptureCollectionViewCell.m
@@ -107,4 +107,14 @@
     }
 }
 
+- (BOOL)isAccessibilityElement
+{
+    return YES;
+}
+
+- (NSString *)accessibilityLabel
+{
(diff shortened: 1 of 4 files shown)


bagia:
I might have missed the Camera capture image. It may need its own Accessibility code. Is it possible to get this cell in the simulator? I couldn't get it. Is it still a thing, or is it something deprecated?

SergioEstevao:
Good work @bagia, I'm going to merge this in!

PR #88, opened 2015-11-19
Issue/73 fix crash when adding photo

### LINKED ISSUE #73: App crashed after photo taken
I debug it at:
https://github.com/wordpress-mobile/MediaPicker-iOS/blob/7a7e99a2c817cb33cf506a142f6387e510e54b66/Pod/Classes/WPMediaCollectionViewController.m#L643

[self.collectionView insertItemsAtIndexPaths:@[[NSIndexPath indexPathForItem:insertPosition inSection:0]]];

Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Invalid update: invalid number of items in section 0.  The number of items contained in an existing section after the update (336) must be equal to the number of items contained in that section before the update (336), plus or minus the number of items inserted or deleted from that section (1 inserted, 0 deleted) and plus or minus the number of items moved into or out of that section (0 moved in, 0 moved out).'

SergioEstevao:
Closes #73 

This fix avoids animate and crash the app on specific cases where changes events arrive before the animation is done.

### DIFF
--- a/Pod/Classes/WPMediaCollectionViewController.m
+++ b/Pod/Classes/WPMediaCollectionViewController.m
@@ -657,13 +657,15 @@ static NSTimeInterval TimeToIgnoreNotificationAfterAddition = 2;
 
 - (void)processMediaCaptured:(NSDictionary *)info
 {
+    NSInteger mediaItemsBefore = [self.dataSource numberOfAssets];
     self.ignoreMediaTimestamp = [NSDate timeIntervalSinceReferenceDate];
     WPMediaAddedBlock completionBlock = ^(id<WPMediaAsset> media, NSError *error) {
         if (error || !media) {
             NSLog(@"%@", error);
             return;
         }
-        [self addMedia:media];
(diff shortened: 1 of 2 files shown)


Pod/Classes/WPMediaCollectionViewController.m
```diff
@@ -657,13 +660,15 @@ - (void)showAlertAboutMediaCapturePermission
 
 - (void)processMediaCaptured:(NSDictionary *)info
 {
+    NSInteger mediaItemsBefore = [self.dataSource numberOfAssets];
     self.ignoreMediaTimestamp = [NSDate timeIntervalSinceReferenceDate];
     WPMediaAddedBlock completionBlock = ^(id<WPMediaAsset> media, NSError *error) {
         if (error || !media) {
-            NSLog(@"%@", error);
+            NSLog(@"Adding media failed: %@", [error localizedDescription]);
```
astralbodies: We shouldn't have any NSLog statements - use DDLogError so they show up in the contain app's logs.

PR #92, opened 2016-01-22
Upgrade to iOS 8 and above

SergioEstevao:
Upgrade the library to iOS 8 and above.
Removed iOS 7 specific code and remove all AssetLibrary code that is deprecated on iOS 8 and above.

How to test:

Open the the sample project and test on several devices on different orientations the picker and specially the showing of alerts for permissions.

Needs Review: @astralbodies


astralbodies:
@SergioEstevao Looks good to go! :shipit:

PR #93, opened 2016-01-27
Remove deprecated rotation methods.

SergioEstevao:
Remove deprecate rotation methods that where missing from the ios8 upgrade.

The code inside the methods was not relevant anymore because it was trying to compensate some rotations issues with the layout code. This code was changed on the iOS 8 upgrade so this fix is no longer needed.


astralbodies:
:shipit: !

PR #99, opened 2016-02-10
Only scroll to first cell when changing group, not on every refresh.

### LINKED ISSUE #98: The picker scroll back to the top when something is changed
When trying to add an image with the mediapicker on my ipad it keeps jumping to the top. When scrolling down images is first a little blurry and then at the moment they get sharp the picker jumps to the top.

Using iCloud photo library with about 12k images and videos

SergioEstevao:
Closes #98 

How to test:
 - Start the sample app
 - Open the picker and scroll  a bit down
 - Switch to the Photos app and delete some photos
 - Go back to sample app and check if the scroll is in the same area.

Needs Review: @bummytime


bummytime:
Works great @SergioEstevao.

:shipit:

PR #101, opened 2016-02-23
Create specific case for video or images.

SergioEstevao:
Add a extra option to show only Images and Videos, and avoid displaying and selection of audio files and other documents.

Needs Review: @koke


koke:
Not sure if the changes to `AppIcon.appiconset` were intentional, but other than that :shipit:

PR #102, opened 2016-02-23
Make capture cell an header instead of a cell.

SergioEstevao:
This allow a better design and simplifies the code around image addition updates and refresh.

Needs Review: @frosty

### DIFF
--- a/Pod/Classes/WPMediaCaptureCollectionViewCell.h
+++ b/Pod/Classes/WPMediaCaptureCollectionViewCell.h
@@ -1,6 +1,6 @@
 @import UIKit;
 
-@interface WPMediaCaptureCollectionViewCell : UICollectionViewCell
+@interface WPMediaCaptureCollectionViewCell : UICollectionReusableView
 
 - (void)stopCaptureOnCompletion:(void (^)(void))block;
 - (void)startCapture;
--- a/Pod/Classes/WPMediaCollectionViewController.m
+++ b/Pod/Classes/WPMediaCollectionViewController.m
@@ -1,6 +1,6 @@
 #import "WPMediaCollectionViewController.h"
 #import "WPMediaCollectionViewCell.h"
-#import "WPMediaCaptureCollectionViewCell.h"
+#import "WPMediaCapturePreviewCollectionView.h"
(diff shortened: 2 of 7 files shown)


Pod/Classes/WPMediaCaptureCollectionViewCell.h
```diff
@@ -1,6 +1,6 @@
 @import UIKit;
 
-@interface WPMediaCaptureCollectionViewCell : UICollectionViewCell
+@interface WPMediaCaptureCollectionViewCell : UICollectionReusableView
```
frosty: Should this be named something other than `Cell`, now it's no longer a cell subclass? `WPMediaCaptureCollectionViewHeader` maybe? ``WPMediaCaptureCollectionViewHeaderView`? ``WPMediaCaptureCollectionViewHeaderFooterView`? :)

Pod/Classes/WPMediaCollectionViewController.m
```diff
@@ -351,6 +329,54 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cell
...
+        return CameraPreviewSize;
+    }
+    return CGSizeZero;
+}
+
+- (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
+           viewForSupplementaryElementOfKind:(NSString *)kind
+                                 atIndexPath:(NSIndexPath *)indexPath
+{
+    if ((kind == UICollectionElementKindSectionHeader && self.showMostRecentFirst) ||
```
frosty: I don't think you need the `if` statement here. This should only ever be hit if you have the capture cell enabled, and you return the same view regardless of whether it's the header or the footer? Also, the documentation says you should never return `nil` from this method.

SergioEstevao: I tested this and it looks that it's only called for non zero sized header and footers, but I will prefer to keep the if to better future proof the code in case of new layouts or adding new headers/footers

SergioEstevao:
@frosty renamed the capture view and comment on the other review comment. Ready for another round of review.

frosty:
Looks good! :shipit:

PR #103, opened 2016-03-03
Improve changes observing

SergioEstevao:
This PR changes the interface for the data source observers to allow reporting of incremental changes.

== How to test ==

To test this properly several scenarios must be run:

 - Add images and videos inside the picker and see if all updates correctly
 - Switch to the Photos app and remove some images and see if it updates correctly when switching back
 - Add some photos using the camera app and then switch back to see if they show
 - Using a iCloud folder test if removal and insertions done remotely (using the iCloud web interface) are done correctly. 
 - Scroll in a big iCloud library and check if no flickering occurs.

Needs Review: @frosty

### DIFF
--- a/Pod/Classes/WPMediaCollectionViewCell.m
+++ b/Pod/Classes/WPMediaCollectionViewCell.m
@@ -1,6 +1,6 @@
 #import "WPMediaCollectionViewCell.h"
 
-static const NSTimeInterval ThredsholdForAnimation = 0.03;
+static const NSTimeInterval ThresholdForAnimation = 0.03;
 static const CGFloat TimeForFadeAnimation = 0.3;
 
 @interface WPMediaCollectionViewCell ()
@@ -96,7 +96,7 @@ static const CGFloat TimeForFadeAnimation = 0.3;
         if (requestKey != self.tag) {
             return;
         }
-        BOOL animated = ([NSDate timeIntervalSinceReferenceDate] - timestamp) > ThredsholdForAnimation;
+        BOOL animated = ([NSDate timeIntervalSinceReferenceDate] - timestamp) > ThresholdForAnimation;
         if ([NSThread isMainThread]){
             [self setImage:result
                   animated:animated];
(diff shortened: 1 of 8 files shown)


astralbodies:
Tested with a ~110GB photo library on an iPhone 6s Plus.
* Changes to the library made remotely didn't show up while the picker was on the screen. I had to put the app into the background and bring it back to show the changes. Not a huge deal but I wonder what would happen if someone selected a deleted image (very small edge case).
* Scrolling seemed fine - should I also test on an iPad?
* Can you provide more details regarding "Using a iCloud archive test if removal and insertions done remotely are done correctly." - I've never used an iCloud archive before.
* Adding & removing from photos app on the device seems to work when coming back to the media picker demo app.

SergioEstevao:
@astralbodies 

iPad testing would be nice.

```Can you provide more details regarding "Using a iCloud archive test if removal and insertions done remotely are done correctly." - I've never used an iCloud archive before```

I was referring to change an iCloud synced folder using the web interface, does that make more sense? :)

File: Pod/Classes/WPMediaCollectionViewCell.m [hunk 1]
```diff
@@ -88,16 +88,15 @@ - (void)setAsset:(id<WPMediaAsset>)asset {
...
         if (error) {
             self.image = nil;
-            NSLog(@"%@", [error localizedDescription]);
             return;
         }
         // Did this request changed meanwhile
         if (requestKey != self.tag) {
             return;
         }
+        BOOL animated = ([NSDate timeIntervalSinceReferenceDate] - timestamp) > ThredsholdForAnimation;
```
frosty: Small typo here (not introduced in this PR, but nice to fix!): `ThredsholdForAnimation`

frosty:
Tested this on my (super slow) iPad 3, and it seems really good. Performance is much better than previous, and scrolling is relatively smooth. I haven't seen any instances of all the photos flashing as new photos load in (which happened previously).

I saw one occasion where the network activity indicator in the status bar continued to spin after dismissing the picker popover, but I've been unable to recreate it. 

I also tested adding / removing photos on another device, and had the same experience as Aaron – I had to background and resume the app before the changes appeared. I don't think that's an issue with this PR though?

I think this is good to :shipit:

PR #104, opened 2016-03-29
Adds Carthage support

koke:
Reorganized project so it's not as coupled to Cocoapods:

- Replaced Xcode project with a top level one with example and framework targets.
- Reorganized file structure so it's less Cocoapods specific.
- Removed empty tests target.
- Add some missing imports.
- Tweak resource loader so it uses the right bundle when it's a framework.

To test:

- Run `carthage build --no-skip-current`. Inspect the resulting framework in `Carthage/Build`
- Test the Example project
- Point WPiOS to this branch and run `pod install`

Needs Review: @SergioEstevao


SergioEstevao:
@koke With this approach can I still test pod locally? using Cocoapods? Before you could run pod install on the Example folder and test the setup of the pod, now it looks it's no longer possible.

SergioEstevao:
I'm also seeing this error when trying to run the carthage command:

```xcodebuild: error: Unable to read project 'WPMediaPicker.xcodeproj' from folder '/Users/sergioestevao/Documents/Automattic/WPMediaPicker/Example'.
	Reason: Project /Users/sergioestevao/Documents/Automattic/WPMediaPicker/Example/WPMediaPicker.xcodeproj cannot be opened because it is missing its project.pbxproj file.```

SergioEstevao:
Is there a concept on carthage for minimum iOS version? on the new MediaPicker project file I see that it's set to 9.3 but this code only needs a minimum of iOS 8.0.

koke:
> With this approach can I still test pod locally? 

I don't think so. This is a shortcoming of Carthage which will try to build any project inside the repo. Since we can't tell it to ignore Pod targets I had to delete them.

> I'm also seeing this error

Try deleting `Example/WPMediaPicker.xcodeproj `. The project was meant to be deleted but you probably have some gitignored user data that's keeping it there

> Is there a concept on carthage for minimum iOS version?

Not specific to carthage, but added it in cf69b2d

koke:
Closing this as it requires more discussion and we're not moving forward with Carthage for now

PR #105, opened 2016-03-29
Alternative carthage implementation

SergioEstevao:
Why no just add a new framework target to the original project.

Needs Review: @koke


koke:
Hmm, I think I was getting confused with another component where Carthage was definitely not picking the right target until I removed all the old ones.

This seems to work. However, I don't like that the target to build the framework is inside an "Example" project.

I think it's a good idea to keep a project where Cocoapods is used to test the pod installation, but maybe it should be a secondary project and not the main one. On the other hand, maybe we should focus on developing/testing in an agnostic way and test the integration with Cocoapods/Carthage via CI.

koke:
Closing this as it requires more discussion and we're not moving forward with Carthage for now

PR #107, opened 2016-03-30
Post-Processing API

jleandroperez:
#### Details:
In this PR we're exposing a new public method, that grants access to the internal `UINavigationController`. This is required to wire a post-processing step, such as cropping.

Needs Review: @SergioEstevao 
Thanks in advance sir!

### DIFF
--- a/Pod/Classes/WPMediaPickerViewController.m
+++ b/Pod/Classes/WPMediaPickerViewController.m
@@ -3,7 +3,7 @@
 #import "WPPHAssetDataSource.h"
 
 @interface WPMediaPickerViewController () <UINavigationControllerDelegate>
-
+@property (nonatomic, strong) UINavigationController *internalNavigationController;
 @end
 
 @implementation WPMediaPickerViewController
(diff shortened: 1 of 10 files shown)


Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -3,7 +3,7 @@
 #import "WPPHAssetDataSource.h"
 
 @interface WPMediaPickerViewController () <UINavigationControllerDelegate>
-
+@property (nonatomic, strong) UINavigationController *internalNavigationController;
```
SergioEstevao: I think this could just be called navigationController no need for internal.

SergioEstevao:
@jleandroperez don't you want to make the change for not dismissing the view controller when doing a single selection?

Do you mind to add a sample of this functionality on the example app? It can just a dummy view controller that show the image in order to be tested there.

Pod/Classes/WPMediaPickerViewController.m
jleandroperez: @SergioEstevao `navigationController` would actually collide with `UIViewController`'s **navigationController** property.

jleandroperez:
@SergioEstevao ready for another look sir!. Added `showDoneButton` property (suggestions for a proper name would be more than welcome!)

SergioEstevao:
:shipit:

jleandroperez:
Yay, thanks for the review @SergioEstevao!!

PR #110, opened 2016-05-31
Add option to use front camera for capture

koke:
If the media picker allows media capturing, this adds the option to use the front camera by default when possible.


SergioEstevao:
:shipit: , thanks for the addition @koke

PR #111, opened 2016-06-23
Fixes for iOS 10

SergioEstevao:
This PR fixes the following issues found on iOS 10:
 - multiple selection not working
 - permissions info plist key on demo app weren't set

How to test:
 - Run the demo app on a iOS 10 device or simulator
 - Check if multiple selection work
 - Test on a device to see if access to camera and video are working

Needs Review: @frosty


frosty:
I've just tested this out on my iPad and multiple selection still doesn't seem to be working correctly. If I tap different cells, the selection moves around. If I tap the same cell repeatedly, then the number on it increments:

multiple-selection

SergioEstevao:
@frosty I'm unable to reproduce the issues that you saw in your iPad. Is this your iPad Pro running iOS 10?

I was able to reproduce the video orientation problem that you where seeing on the camera preview and fixed that.

Can you please give it another look?

frosty:
Okay, I re-tested this on a device (had to delete and clean everything) and it looks good to me! Rotation and selection are working properly.

:shipit:

Issue #112, opened 2016-07-05
Filter of media is not being considered on Groups/Collections


SergioEstevao:
When a filter for media type is active, like just showing videos or just show images, the collection/group and counters are not taking in account the filter type.

This also allows to show groups with zero assets of the selected filter.

PR #117, opened 2016-07-11
Filter poster image of group

SergioEstevao:
Refs #112 

This PR fix the poster image of a group by filtering it according to the media type selected.

How to test:
 - Launch the demo app
 - Tap on options
 - Select filter by video
 - Tap on the plus
 - Tap on the Group selector on the top
 - Check if the poster images show for each group reflect a video asset in the group.

Needs Review: @frosty

### DIFF
--- a/Pod/Classes/WPPHAssetDataSource.h
+++ b/Pod/Classes/WPPHAssetDataSource.h
@@ -20,6 +20,11 @@ NS_CLASS_AVAILABLE_IOS(8_0) @interface WPPHAssetDataSource : NSObject<WPMediaCol
 /**
  An implementation of the WPMediaGroup protocol using the PHAssetCollection class
  */
-@interface PHAssetCollection(WPMediaGroup)<WPMediaGroup>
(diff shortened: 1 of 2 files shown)


Pod/Classes/WPPHAssetDataSource.h
```diff
@@ -20,6 +20,11 @@ NS_CLASS_AVAILABLE_IOS(8_0) @interface WPPHAssetDataSource : NSObject<WPMediaCol
 /**
  An implementation of the WPMediaGroup protocol using the PHAssetCollection class
  */
-@interface PHAssetCollection(WPMediaGroup)<WPMediaGroup>
+@interface PHAssetCollectionForWPMediaGroup:NSObject <WPMediaGroup>
```
frosty: Totally nitpicky, but I'd prefer some spaces around that `:` :)

frosty:
Works for me! 👍 One tiny code comment, otherwise...

:shipit:

PR #130, opened 2017-01-06
Specifies a less specific session quality to prevent iPhone 4s crash.

### LINKED ISSUE #129: Crash with iPhone 4s due to camera configuration
The media picker crashes consistently on an iPhone 4s due to the front-facing camera not supporting 1280x720.

```
Fatal Exception: NSInvalidArgumentException
*** Can't add <AVCaptureDeviceInput: 0x1adf7fe0 [Front Camera]> because the device does not support AVCaptureSessionPreset1280x720. Use -[AVCaptureDevice supportsAVCaptureSessionPreset:].
```

Prevent the crash by maybe inspecting the device to see capabilities or by using one of the high/medium/low presets rather than a specific resolution.

cc: @SergioEstevao

astralbodies:
Closes #129 

Using this table: http://stackoverflow.com/a/31964333

Setting the session capture quality to `AVCaptureSessionPresetHigh` rather than `AVCaptureSessionPreset1280x720` which isn't supported on the front camera for the iPhone 4s.

@kurzee did verify this fixed the crash.

### DIFF
--- a/Pod/Classes/WPMediaCapturePreviewCollectionView.m
+++ b/Pod/Classes/WPMediaCapturePreviewCollectionView.m
@@ -79,7 +79,7 @@
 
         if (!self.session){
             self.session = [[AVCaptureSession alloc] init];
-            self.session.sessionPreset = AVCaptureSessionPreset1280x720;
+            self.session.sessionPreset = AVCaptureSessionPresetHigh;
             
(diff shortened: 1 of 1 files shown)


SergioEstevao:
:shipit:

PR #132, opened 2017-01-09
Tweaks data source so sort order matches Photos app.

### LINKED ISSUE #131: Photos in Camera Roll and other albums might not match order in Photo app
As reported:
```
Today I ran into this issue:

    AirDropped a photo originally taken in September 2016 to my phone.
    It showed up as the most recent photo in the Photos app.
    Posted to Instagram. It appeared in the photo library without issue.
    Went to post to my test site, where I repost my Instagrams by hand for testing the WP mobile app.
    Looked in “All Photos” but couldn’t find the photo.
    After a bit of searching, finally realized it was appearing pretty far into the scrollback, ordered by the original date on the photo (i.e. different ordering than what I’m seeing in the Photos app or Instagram).

Probably pretty edge case but just wanted to note it, as it was pretty confusing.
```

I was able to reproduce the issue by following the above steps.  

It seems that the Photos app does not sort PHAssets by creationDate or modificationDate, but by when the photo was added to the collection/album.  I can match the ordering of the Photos app by passing `nil` instead of a `PHFetchOptions` instance when calling `[PHAsset fetchAssetsInAssetCollection:options]` but we want the inverse of this ordering.   We could probably manage the correct sort via one or more helper methods to [adjust the row index](https://stackoverflow.com/a/39645403).

aerych:
Fixes #131 

This PR removes the sortDescriptor passed when fetching assets and instead updates the `mediaAtIndex` helper method to return results in either ascending or descending order.  

This resolves an issue where sorting on either `creationDate` or `modificationDate` would case some photos to be ordered differently than their ordering in the iOS Photos app.

To test:

Follow the steps out lined in the issue and confirm that the photos match the order shown in the Photos app.

Needs review: @SergioEstevao

### DIFF
--- a/Pod/Classes/WPPHAssetDataSource.m
+++ b/Pod/Classes/WPPHAssetDataSource.m
@@ -226,17 +226,23 @@
 
 - (id<WPMediaAsset>)mediaAtIndex:(NSInteger)index
 {
+    NSInteger count = [self numberOfAssets];
+    if (count == 0) {
+        return nil;
+    }
+
     if (self.ascendingOrdering) {
+        if (index < 0 || index >= count) {
+            return nil;
+        }
         return self.assets[index];
     }
 
     // Adjust the index so items are returned in reverse order.
     // We do this, rather than specifying the sort order in PHFetchOptions,
     // to preserve the sort order of assets in the Photos app (only in reverse).
-    NSInteger count = [self numberOfAssets];
     NSInteger idx = (count - 1) - index;
-
-    if (count == 0 || idx < 0 || idx >= count ) {
+    if (idx < 0 || idx >= count ) {
         return nil;
     }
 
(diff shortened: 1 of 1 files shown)


SergioEstevao:
@aerych did you see my comment above?

aerych:
@SergioEstevao Yeppers!  I've been side tracked working on a few other things.  Hoping to switch back to this either later today or early tomorrow. :)

aerych:
@SergioEstevao I think this is ready for another round.  I've adapted the patch to also adjust indexes for deletes, moves, and inserts.  To test I did the following:
- Launched the demo app, 
- Opened the collection view 
- Backgrounded the app.
- Opened Safari and performed a google images search for "Coffee".
- I tapped on several images and long pressed to save the images.
- I opened the Photos app and deleted multiple images. 
- Opened the demo app and confirmed that A) there were no crashes and B) the photos shown were correctly updated to match the photos in the Photo app. 

I iterated on the above steps in different ways:
- just adding photos
- just deleting photos
- deleting then adding photos
- adding then deleting photos

Everything seemed to work as intended (fingers crossed!)

SergioEstevao:
:shipit: good work @aerych !

aerych:
Thanks @SergioEstevao !!

PR #133, opened 2017-01-12
Release/0.10.4

aerych:
Bumps the podspec version and updates the example.

@SergioEstevao its a simple patch but would you still take a peak for paranoia sake?


aerych:
Closing in favor of v 0.11

PR #137, opened 2017-02-08
Fix pod spec lint warnings.

SergioEstevao:
This PR fix some warning that the build project was giving and making the pod spec lint complain.


.swift-version
```diff
@@ -0,0 +1 @@
+2.3
```
koke: I believe the swift version message is [only displayed as a hint when there's any warning](https://github.com/CocoaPods/CocoaPods/blob/cae98c5386f7d80d6fe065c3f081ce130628151a/lib/cocoapods/validator.rb#L146) and if you fix the rest of the warnings, it'll go away

SergioEstevao:
@koke I basically run Xcode build locally and check if the warning went away.
Pod spec lint does the same.

koke:
OK, can you drop the `.swift-version` and see if it works? If it has to be there, I'd use 3.0, as 2.3 will probably be removed in a future Xcode.

SergioEstevao:
@koke you can test now, I pushed a tag with the new version to repo. It's not published on Cocoapods but it allows to test using ```pod spec lint```

PR #138, opened 2017-02-16
Issue/refactor default data source

SergioEstevao:
This PR just moves the default data source from outside the picker. 

This is some preparatory work to make the implementation of a custom view for the picker easier.

How to test:

 - Run the demo app
 - And check the picker opens correctly.


Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -45,7 +45,7 @@ - (void)setupNavigationController
     vc.filter = self.filter;
     vc.allowMultipleSelection = self.allowMultipleSelection;
     if (!self.dataSource) {
-        self.dataSource = [self defaulDataSource];
+        self.dataSource = [WPPHAssetDataSource sharedInstance];
```
frosty: Does this need to be a singleton? Can we not just create a new one? I guess the issue is that `dataSource` is a weak reference?

SergioEstevao: Correct.

SergioEstevao: And keep in this this only applied has a default to allow the picker to work without extra configuration.

frosty:
Sorry for the delay – :shipit:!

PR #141, opened 2017-02-24
Issue/140 fix crash when editing albums

### LINKED ISSUE #140: Crash when editing albums on the Photos app with picker open.
Someone can rewrite it in swift?

SergioEstevao:
Fixes #140 

This PR fixes a crash that was happening when album changes where made in the Photos.app when the album picker was open in the MediaPicker.
It also make sure the album list gets updated correctly when changes are made.

How to test:

 - Open the Media Picker demo app and start the picker
 - Tap the group selector on the top
 - While keeping it open, switch to the Photos app
 - Create/Delete/Edit a user album
 - Switch back to the demo app
 - Check if no crash happens and the albums are update accordingly.


frosty:
Looks good! :shipit:

PR #143, opened 2017-03-01
Crash fix: dispatch entire photoLibraryDidChange callback to main queue

### LINKED ISSUE #128: Crash in WPMediaCollectionViewController updateDataWithRemoved:inserted:changed:moved
We've had quite a few reports for a crash affecting the media picker – something going wrong with the collection view in `-[WPMediaCollectionViewController updateDataWithRemoved:inserted:changed:moved:]`. 

screen shot 2016-11-16 at 14 56 32

Possibly-related exceptions:

> Fatal Exception: NSInternalInconsistencyException
attempt to insert item 0 into section 0, but there are only 0 items in section 0 after the update

> Fatal Exception: NSInternalInconsistencyException
Invalid update: invalid number of items in section 0. The number of items contained in an existing section after the update (2063) must be equal to the number of items contained in that section before the update (2063), plus or minus the number of items inserted or deleted from that section (1 inserted, 0 deleted) and plus or minus the number of items moved into or out of that section (0 moved in, 0 moved out)

kurzee:
Fixes #128 

This should resolve another recurring crash for Media.

Looking at the documentation for the `photoLibraryDidChange` used in `WPPHAssetDataSource`, you can see Apple's example of wrapping the entire block to dispatch to the main queue, here:
- https://developer.apple.com/reference/photos/phphotolibrarychangeobserver?language=objc

Without this fix, I suspect there was a case in which the background queue ends up firing `photoLibraryDidChange`, running the changes in that same background queue, but following up with an async UI update that landed out of sync with the model changes. Wrapping the whole block should prevent this. 🤞 

Needs review: @frosty, @SergioEstevao or @astralbodies (we should put this into 7.1 as well)

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,15 +1,15 @@
 PODS:
-  - WPMediaPicker (0.11.2)
+  - WPMediaPicker (0.11.3)
 
 DEPENDENCIES:
   - WPMediaPicker (from `../`)
 
 EXTERNAL SOURCES:
   WPMediaPicker:
-    :path: ../
+    :path: "../"
 
 SPEC CHECKSUMS:
-  WPMediaPicker: 8504d1547b41e5cb024de97d51e2c6b56d8cded7
+  WPMediaPicker: 24a218bcd70cc9620c894129d8bc9d5e1935bd23
 
 PODFILE CHECKSUM: 7855568785f801c5559f8e70856ad87de227dc95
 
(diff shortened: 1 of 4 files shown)


SergioEstevao:
@kurzee I agree with the changes except with the bump to the pod spec version on this PR.

That should be done on separate PR just to create the new release.

kurzee:
>except with the bump to the pod spec version on this PR. ... That should be done on separate PR just to create the new release.

👍 After review, I'll close this one then reopen and merge with only the code changes. Plus open up a separate release PR.

kurzee:
Merging this fix along in https://github.com/wordpress-mobile/MediaPicker-iOS/pull/144

Merging the release alone in https://github.com/wordpress-mobile/MediaPicker-iOS/pull/145

PR #147, opened 2017-03-06
Refactor class structure to allow reuse of the media picker collection view.

SergioEstevao:
This refactor will allow better reuse of the core functional of the picker that is the collection view.


frosty:
This seems okay to me! Obviously needs some tweaks in WPiOS, but with those in place things seem to be functioning as expected. :shipit:

PR #150, opened 2017-03-07
Implement a Media Picker Input View 

SergioEstevao:
This PR implements a view integrating a mediaPicker that can be used as an inputView.

How to test:
 - Launch the demo app
 - Tap on the input field to on the top
 - Check if custom input view shows up
 - Select media on it
 - Tap Done
 - See if media shows up in the list
 - Play around with the media options to see if the scenarios work ok.
 - Push it to the limit.


astralbodies:
There is no Album selection in the image picker - I had picked an album other than All Photos in the full-screen picker, closed out, then tapped into the text box. I then took a picture and because the picture didn't show up in that album I had no way of inserting that picture into the text box.

astralbodies:
Taking a picture also created a noisy array of errors:

```2017-03-10 15:12:16.108847 WPMediaPicker[11263:4696771] [Camera] [CAMCaptureEngine] Received a session runtime error notification  with error: Error Domain=AVFoundationErrorDomain Code=-11800 "The operation could not be completed" UserInfo={NSUnderlyingError=0x170257c40 {Error Domain=NSOSStatusErrorDomain Code=-12785 "(null)"}, NSLocalizedFailureReason=An unknown error occurred (-12785), NSLocalizedDescription=The operation could not be completed}
2017-03-10 15:12:16.109006 WPMediaPicker[11263:4696771] [Camera] [CAMCaptureEngine] Performing recovery from error: Error Domain=AVFoundationErrorDomain Code=-11800 "The operation could not be completed" UserInfo={NSUnderlyingError=0x170257c40 {Error Domain=NSOSStatusErrorDomain Code=-12785 "(null)"}, NSLocalizedFailureReason=An unknown error occurred (-12785), NSLocalizedDescription=The operation could not be completed}
2017-03-10 15:12:16.109196 WPMediaPicker[11263:4696771] [Camera] [CAMCaptureEngine] Attempting to recover from a session runtime error by restarting the AVCaptureSession...
2017-03-10 15:12:16.114245 WPMediaPicker[11263:4696771] [Camera] [CAMCaptureEngine] Received a session runtime error notification during recovery  with error: Error Domain=AVFoundationErrorDomain Code=-11800 "The operation could not be completed" UserInfo={NSUnderlyingError=0x170454820 {Error Domain=NSOSStatusErrorDomain Code=-12785 "(null)"}, NSLocalizedFailureReason=An unknown error occurred (-12785), NSLocalizedDescription=The operation could not be completed}
2017-03-10 15:12:16.114346 WPMediaPicker[11263:4696771] [Camera] [CAMCaptureEngine] Encountered another AVCaptureSession runtime error during session recovery… (Error Domain=AVFoundationErrorDomain Code=-11800 "The operation could not be completed" UserInfo={NSUnderlyingError=0x170454820 {Error Domain=NSOSStatusErrorDomain Code=-12785 "(null)"}, NSLocalizedFailureReason=An unknown error occurred (-12785), NSLocalizedDescription=The operation could not be completed})
2017-03-10 15:12:42.527909 WPMediaPicker[11263:4696771] Adding media failed: (null)
```

astralbodies:
Tapping a second time into the text box makes the media picker untappable/hang for a while.

```2017-03-10 15:16:34.649152 WPMediaPicker[11263:4696771] Presenting view controllers on detached view controllers is discouraged <WPMediaPickerViewController: 0x100416b60>.
2017-03-10 15:16:53.463496 WPMediaPicker[11263:4696771] [App] if we're in the real pre-commit handler we can't actually add any new fences due to CA restriction
2017-03-10 15:16:53.463669 WPMediaPicker[11263:4696771] [App] if we're in the real pre-commit handler we can't actually add any new fences due to CA restriction
2017-03-10 15:16:54.451437 WPMediaPicker[11263:4696771] [App] if we're in the real pre-commit handler we can't actually add any new fences due to CA restriction
2017-03-10 15:16:54.451562 WPMediaPicker[11263:4696771] [App] if we're in the real pre-commit handler we can't actually add any new fences due to CA restriction
```

astralbodies:
Seems like the media picker should return to the first entry in the list when it's brought back up. If I tap back in the field it starts at the last scroll position I was at when it was last dismissed.

Pod/Classes/WPInputMediaPickerView.m
```diff
@@ -0,0 +1,87 @@
...
+- (instancetype)initWithCoder:(NSCoder *)aDecoder {
+    self = [super initWithCoder: aDecoder];
+    if (self) {
+        [self commonInit];
+    }
+    return self;
+}
+
+- (void)commonInit {
+    WPMediaPickerViewController *vc = [[WPMediaPickerViewController alloc] init];
```
astralbodies: Containing a view controller inside of a view violates the containment policy of view controllers (addChild, etc). It seems like rotation works properly but there's a lot that goes on under the hood with VCs you're not getting doing it this way.

SergioEstevao:
@astralbodies thanks for the review, taking in account your comments I think the way forward will be the one of the following:

1) Create a proper picker view
 - Create a new view that extends from collection view, but implements all the logic for the media picker
 - That way we can use that new view directly for the input view

2) Create a View Controller called InputViewController and that creates a view that encapsulates the current MediaPickerViewController.

What do you think?

SergioEstevao:
@astralbodies: I don't think the input view should have an album selector, this should be a way to add some quick photo from the Camera Roll, if you want to select albums we should create a link to the full screen media picker.

If you see the apps, out there that have a an input media picker, they normally just have the camera roll. Check iMessages, Messages from Facebook, Telegram.

SergioEstevao:
@astralbodies ready for another look.

I took the option 2 from the comment above and refactor the code to use a ViewController to wrap the input view.

This made the code a lot simpler a clear and solve some presentation issues.

SergioEstevao:
Just wanted to add some extra clarification about the purpose of this PR:

 - The solution been created here for media picking in the input view is not in anyway the solution that we will need to use in the WordPress app
 - This PR served has a test for an possible input view implementation, and what where the challenges behind it, and what possible architecture changes we may need to do in order to support it.
 - I think it's also a good example implementation of the a custom input view for users/developers of the library that are not using inside the WordPress eco-system.
- And has a final point I think I wanted to have this merged in order to experience on Aztec for the media flow of adding images in the editor.

astralbodies:
Looks like the changes are good; I don't see any of the hangs and errors so we're good to merge.

PR #151, opened 2017-03-07
Fixes nullability warnings in WPMediaPickerViewController.

frosty:
I think our recent refactoring introduced a few nullability warnings which I was seeing once pulling the library into WPiOS. This should fix them.

screen shot 2017-03-07 at 17 52 09

To test:

* Check I've used the correct nullability types
* Update the Podfile in a local copy of WPiOS to point to this branch and ensure that you don't see any nullability warnings when building.


frosty:
Thanks @SergioEstevao!

PR #154, opened 2017-03-10
Tweaks required for Media Library deletion in WPiOS

frosty:
This PR makes a few tweaks / additions that are required to support the Media Library deletion PR for https://github.com/wordpress-mobile/WordPress-iOS/pull/6858. 

**Changes:**

* If the custom preview view controller delegate method `mediaPickerController:previewViewControllerForAsset:` is implemented but returns `nil`, we don't show a custom preview. Previously we'd show the default one, but now you can do that just by not implementing this delegate method.
* If `allowsMultipleSelection` is changed after the collection view is instantiated, we now update the collection view's matching property. Previously, you could only set this once.
* Added a method for clearing the current selection in the collection view.

### DIFF
--- a/Pod/Classes/WPMediaPickerViewController.h
+++ b/Pod/Classes/WPMediaPickerViewController.h
@@ -165,5 +165,10 @@
  */
 - (void)setGroup:(nonnull id<WPMediaGroup>)group;
 
+/**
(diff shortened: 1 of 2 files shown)


SergioEstevao:
:shipit:

PR #155, opened 2017-03-14
Issue/135: Don't silently pass through face up/down device orientations

### LINKED ISSUE #135: Camera preview orientation incorrect when tilted forward
When holding my phone in portrait orientation, if I tilt it forward past about 45 degrees the camera preview header flips into landscape orientation. I did this when getting ready to take a picture pointing straight down at my desk. It's worth noting that both my phone and the camera app's orientations were still portrait at this point.

Tested: Latest WPiOS 6.9 beta build, iPhone 7, iOS 10.2.1

frosty:
Fixes #135.

The camera preview image was getting flipped when holding the device in portrait and then tilting it forward (as though you were going to point it downwards). This seemed to be because we were casting `UIDeviceOrientation` directly to `AVCaptureVideoOrientation`, but `AVCaptureVideoOrientation` has two extra Face Up / Face Down cases that the AV enum doesn't have. To fix the issue, we convert either of those cases to `AVCaptureVideoOrientationPortrait`.

To test:

* Run on a device
* Open the picker and look at the camera preview cell
* Hold the device in portrait
* Slowly tilt the device forward until you're pointing the camera down at the floor. Check the preview image is still orientated correctly
* Check that landscape orientations still work correctly too.

Needs review: @SergioEstevao

### DIFF
--- a/Pod/Classes/WPMediaCapturePreviewCollectionView.m
+++ b/Pod/Classes/WPMediaCapturePreviewCollectionView.m
@@ -100,7 +100,7 @@
                 self.captureVideoPreviewLayer = [[AVCaptureVideoPreviewLayer alloc] initWithSession:self.session];
                 self.captureVideoPreviewLayer.videoGravity = AVLayerVideoGravityResizeAspectFill;
                 self.captureVideoPreviewLayer.frame = viewLayer.bounds;
-                self.captureVideoPreviewLayer.connection.videoOrientation = [self videoOrientationForDeviceOrientation:[[UIDevice currentDevice] orientation]];
+                self.captureVideoPreviewLayer.connection.videoOrientation = [self videoOrientationForInterfaceOrientation:[[UIApplication sharedApplication] statusBarOrientation]];
                 [viewLayer addSublayer:_captureVideoPreviewLayer];
             });
         }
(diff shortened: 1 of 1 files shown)


Pod/Classes/WPMediaCapturePreviewCollectionView.m
```diff
@@ -110,10 +110,19 @@ - (void)startCapture
...
     }
 }
 
+- (AVCaptureVideoOrientation)videoOrientationForDeviceOrientation:(UIDeviceOrientation)orientation
+{
+    if (orientation == UIDeviceOrientationFaceUp || orientation == UIDeviceOrientationFaceDown) {
+        return AVCaptureVideoOrientationPortrait;
+    }
+
+    return (AVCaptureVideoOrientation)orientation;
```
koke: `AVCaptureVideoOrientation` can't be directly mapped to `UIDeviceOrientation`, as their values are different.

frosty:
Thanks @koke, I hadn't thought about landscape using face down orientation too. I've switched things over to use the interface orientation, as you suggested, and I've explicitly mapped to the correct values.

Better?

frosty:
Thanks @SergioEstevao and @koke!

PR #159, opened 2017-03-23
Video player toolbar.

SergioEstevao:
This PR implement on the video player a toolbar that allows to play/pause the video and to display the current time and total time of a video.

screen shot 2017-03-23 at 12 28 55

To test:
 - Start the demo app
 - Tap on the plus button
 - Force touch/ Long press on a video 
 - Check if in the force touch preview no toolbar shows up
 - Check that if the full screen preview if you tap once in the video the toolbar shows and the value are correct.
 - Tap again on the video and see if the toolbar goes away.
 - Play around with the play/pause button

### DIFF
--- a/Example/WPMediaPicker/DemoViewController.m
+++ b/Example/WPMediaPicker/DemoViewController.m
@@ -184,15 +184,6 @@
     [self.mediaPicker showAfterViewController:postProcessingViewController];
 }
 
-- (UIViewController *)mediaPickerController:(WPMediaPickerViewController *)picker previewViewControllerForAsset:(id<WPMediaAsset>)asset
-{
-    if ([self.options[MediaPickerOptionsCustomPreview] boolValue] == false) {
-        return nil;
-    }
-
-    return [[CustomPreviewViewController alloc] initWithAsset:asset];
-}
-
 #pragma - Actions
 
 - (void) clearSelection:(id) sender
--- a/Example/WPMediaPicker/DemoViewController.m
+++ b/Example/WPMediaPicker/DemoViewController.m
@@ -192,7 +192,10 @@
         return [[CustomPreviewViewController alloc] initWithAsset:asset];;
     }
 
-    return [[WPAssetViewController alloc] initWithAsset: asset];
+    WPAssetViewController *assetViewController = [[WPAssetViewController alloc] initWithAsset: asset];
+    assetViewController.delegate = picker;
+    assetViewController.selected = [picker.selectedAssets containsObject:asset];
+    return assetViewController;
 
 }
 
--- a/Example/WPMediaPicker/DemoViewController.m
+++ b/Example/WPMediaPicker/DemoViewController.m
@@ -189,7 +189,7 @@
         return nil;
     }
     if ([self.options[MediaPickerOptionsCustomPreview] boolValue]) {
-        return [[CustomPreviewViewController alloc] initWithAsset:asset];;
+        return [[CustomPreviewViewController alloc] initWithAsset:asset];
     }
 
     WPAssetViewController *assetViewController = [[WPAssetViewController alloc] initWithAsset: asset];
--- a/Example/WPMediaPicker/DemoViewController.m
+++ b/Example/WPMediaPicker/DemoViewController.m
@@ -184,6 +184,18 @@
     [self.mediaPicker showAfterViewController:postProcessingViewController];
 }
 
+- (UIViewController *)mediaPickerController:(WPMediaPickerViewController *)picker previewViewControllerForAsset:(id<WPMediaAsset>)asset {
+    if (picker == self.mediaInputViewController.mediaPicker) {
+        return nil;
+    }
+    if ([self.options[MediaPickerOptionsCustomPreview] boolValue]) {
+        return [[CustomPreviewViewController alloc] initWithAsset:asset];;
+    }
+
+    return [[WPAssetViewController alloc] initWithAsset: asset];
+
+}
+
 #pragma - Actions
 
 - (void) clearSelection:(id) sender
--- a/Pod/Classes/WPAssetViewController.m
+++ b/Pod/Classes/WPAssetViewController.m
@@ -171,8 +171,8 @@
 - (void)handleTapOnAsset:(UIGestureRecognizer *)gestureRecognizer
 {
     if (gestureRecognizer.state == UIGestureRecognizerStateEnded) {
-        [self.navigationController setNavigationBarHidden:!self.navigationController.isNavigationBarHidden animated:YES];
-        [self.videoView setControlToolbarHidden: self.navigationController.isNavigationBarHidden animated: YES];
+        [self.navigationController setNavigationBarHidden:!self.videoView.controlToolbarHidden animated:YES];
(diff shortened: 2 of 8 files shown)


Example/WPMediaPicker/DemoViewController.m
```diff
@@ -184,13 +184,19 @@ - (void)mediaPickerController:(WPMediaPickerViewController *)picker didFinishPic
...
 
-- (UIViewController *)mediaPickerController:(WPMediaPickerViewController *)picker previewViewControllerForAsset:(id<WPMediaAsset>)asset
-{
-    if ([self.options[MediaPickerOptionsCustomPreview] boolValue] == false) {
+- (UIViewController *)mediaPickerController:(WPMediaPickerViewController *)picker previewViewControllerForAsset:(id<WPMediaAsset>)asset {
+    if (picker == self.mediaInputViewController.mediaPicker) {
         return nil;
     }
+    if ([self.options[MediaPickerOptionsCustomPreview] boolValue]) {
+        return [[CustomPreviewViewController alloc] initWithAsset:asset];;
```
frosty: Nitpick, double `;;`.

Example/WPMediaPicker/DemoViewController.m
```diff
@@ -184,13 +184,19 @@ - (void)mediaPickerController:(WPMediaPickerViewController *)picker didFinishPic
...
+    }
+
+    WPAssetViewController *assetViewController = [[WPAssetViewController alloc] initWithAsset: asset];
+    assetViewController.delegate = picker;
+    assetViewController.selected = [picker.selectedAssets containsObject:asset];
+    return assetViewController;
```
frosty: 👍 Thanks for fixing this for the change I made to the delegate!

Pod/Classes/WPAssetViewController.m
```diff
@@ -161,7 +171,8 @@ - (void)showError:(NSError *)error {
 - (void)handleTapOnAsset:(UIGestureRecognizer *)gestureRecognizer
 {
     if (gestureRecognizer.state == UIGestureRecognizerStateEnded) {
-        [self.navigationController setNavigationBarHidden:!self.navigationController.isNavigationBarHidden animated:YES];
+        [self.navigationController setNavigationBarHidden:!self.videoView.controlToolbarHidden animated:YES];
```
frosty: I think we should probably also hide the status bar when we hide the navigation bar / control bar? Otherwise it's overlaid on the video and we don't get 'proper' fullscreen.

Pod/Classes/WPVideoPlayerView.m
```diff
@@ -41,18 +45,54 @@ - (void)commonInit {
...
+    [_player removeTimeObserver:self.timeObserver];
+    [_player pause];
     _asset = nil;
+    _player = nil;
 }
 
 - (void)layoutSubviews {
     [super layoutSubviews];
     self.playerLayer.frame = self.bounds;
+    CGFloat position = toolbarHeight;
```
frosty: Could just do this as a ternary? `CGFloat position = (self.controlToolbarHidden) ? 0 : toolbarHeight;`

Pod/Classes/WPVideoPlayerView.m
```diff
@@ -41,18 +45,54 @@ - (void)commonInit {
...
+    if (self.controlToolbarHidden) {
+        position = 0;
+    }
+    self.controlToolbar.frame = CGRectMake(0, self.frame.size.height - position, self.frame.size.width, toolbarHeight);
+}
+
+- (UIToolbar *)controlToolbar {
+    if (_controlToolbar) {
```
frosty: Total nitpick, but this method and `videoDurationButton` handle an existing ivar differently – this one checks if it exists, `videoDurationButton` checks if it's nil. Can we do it the same way for both? :)

Pod/Classes/WPVideoPlayerView.m
```diff
@@ -41,18 +45,54 @@ - (void)commonInit {
...
+        return _controlToolbar;
+    }
+    _controlToolbar = [[UIToolbar alloc] init];
+    _controlToolbar.hidden = YES;
```
frosty: I was playing around with the styles, and I think things look best if we have a black translucent toolbar, with a white play button:

```
_controlToolbar.tintColor = [UIColor whiteColor];
_controlToolbar.barStyle = UIBarStyleBlack;
_controlToolbar.translucent = YES;
```

Pod/Classes/WPVideoPlayerView.m
```diff
@@ -41,18 +45,54 @@ - (void)commonInit {
...
+    [self updateControlToolbar];
+    return _controlToolbar;
+}
+
+- (UIBarButtonItem *)videoDurationButton {
+    if (_videoDurationButton == nil) {
+        _videoDurationButton = [[UIBarButtonItem alloc] initWithTitle:@"" style:UIBarButtonItemStylePlain target:nil action:nil];
```
frosty: I saw an issue during testing where the button's text shifts slightly to the left or the right each time it updates its value. I just tried out switching things to use a right-aligned label with a fixed width, and it seems to behave much better. There's a diff here you can check out: https://gist.github.com/frosty/1d37e066224dc4b99551bcf4a1d29a49

The only thing I didn't add is updating the label's width on view rotation, which we should probably handle.

What do you think?

Pod/Classes/WPVideoPlayerView.m
```diff
@@ -89,6 +131,89 @@ - (void)playerItemDidReachEnd:(AVPlayerItem *)playerItem {
...
+    }
+}
+
+- (void)setControlToolbarHidden:(BOOL)hidden animated:(BOOL)animated {
+    CGFloat animationDuration = animated ? UINavigationControllerHideShowBarDuration : 0;
+    if (!hidden) {
+        self.controlToolbar.hidden = hidden;
+    }
+    [UIView animateWithDuration:animationDuration animations:^{
+        CGFloat position = self.controlToolbar.frame.size.height;
```
frosty: `CGFloat position = (hidden) ? 0 : self.controlToolbar.frame.size.height;`?

Pod/Classes/WPVideoPlayerView.m
```diff
@@ -89,6 +131,89 @@ - (void)playerItemDidReachEnd:(AVPlayerItem *)playerItem {
...
+    self.videoDurationButton.title = [NSString stringWithFormat:@"%@/%@", currentDuration, totalDuration];
+}
+
+- (NSString *)stringFromTimeInterval:(NSTimeInterval)timeInterval
+{
+    NSInteger roundedHours = floor(timeInterval / 3600);
+    NSInteger roundedMinutes = floor((timeInterval - (3600 * roundedHours)) / 60);
+    NSInteger roundedSeconds = round(timeInterval - (roundedHours * 60 * 60) - (roundedMinutes * 60));
+
+    if (roundedHours > 0)
```
frosty: Could we use braces around these `if` statement clauses?

SergioEstevao:

frosty:
Thanks for the changes, @SergioEstevao! I think the duration label may look better if the button is disabled, but your call.

Other than that, let's :shipit:!

SergioEstevao:
Thanks @frosty for the review and suggestions!

PR #160, opened 2017-03-27
Issue/fix scroll issues

SergioEstevao:
Fixes some scrolling issues that where affecting scroll to start.

To test:
 - Open the picker on the + button and check if the scroll to top or bottom depending of the options selected occur correctly.


frosty:
Looks good to me! :shipit:

PR #168, opened 2017-04-28
Tweak cell layout calculations to avoid large gaps between cells.

frosty:
This PR makes a few tweaks to the cell layout logic of the collection view to avoid large gaps between cells when displayed at certain sizes. After calculating the cell size, we simply check whether all of the available space is filled with cells. If not, we floor the number of cells we can fit into the space and recalculate the cell width based on that.

Before:

simulator screen shot 28 apr 2017 16 14 36

After:

simulator screen shot 28 apr 2017 16 06 13

To test:

* Point WPiOS to this branch and open the media library in landscape. You shouldn't see any large gaps between cells. 
* Check iPhone still looks correct!


frosty:
Thanks Sergio!

PR #169, opened 2017-04-28
Optimize thumbnail fetch for albums

SergioEstevao:
This PR optimizes the fetching of images by using opportunistic fetch of images and caching albums groups.

To test:
 - Open the media picker demo on a device with a big media collection ( > 5K) and with iCloud activated
 - Scroll the items up and down and see if the scrolling is quick and with no pauses.
 - Open the group picker and scroll up down the picker
 - Switch to the photos app and add image and add/remove albums
 - Switch back to the demo app and see if update happen correctly.


Pod/Classes/WPPHAssetDataSource.m
```diff
@@ -541,11 +546,17 @@ - (NSString *)identifier
...
+    NSInteger count = self.collection.estimatedAssetCount;
+    if (count != NSNotFound) {
+        return count;
+    }
+    if (!self.fetchResult){
+        PHFetchOptions *fetchOptions = [PHFetchOptions new];
+        fetchOptions.predicate = [WPPHAssetDataSource predicateForFilterMediaType:mediaType];        
+        self.fetchResult = [PHAsset fetchAssetsInAssetCollection:self.collection options:fetchOptions];
+    }
+    count = self.fetchResult.count;
```
frosty: We could just `return self.fetchResult.count` here

frosty:
Thanks for making these changes! Scrolling the main collection view seems smooth to me.

The groups picker still takes quite a long time to appear when I first tap it, and sometimes freezes up a little. I can make it freeze every time I've scrolled down the the list and then am scrolling back to the top – it gets stuck about 4 items from the top and then jumps to the top. WPiOS does that too though, so I don't think that's anything caused by this PR. Overall, loading the images seems a little faster and smoother which is great :)

One code comment, otherwise :shipit:

PR #170, opened 2017-04-28
Set an default image for asset that fail to produce an thumbnail.

SergioEstevao:
This PR sets an default thumbnail image for video and images when none is available.

The end result at the moment is this:

screen shot 2017-04-28 at 22 43 53

@frosty I was thinking that I probably should replace my current icons for the photo and video for the same gridicons style you got for audio and files. 

How did you generate those images?


frosty:
@SergioEstevao I exported the Gridicons PDFs from Sketch into the appropriate PNG files – just sent you the icons for this PR!

Pod/Classes/WPMediaCollectionViewCell.m
```diff
@@ -193,8 +193,17 @@ - (void)fetchAssetImage
...
+            self.imageView.contentMode = UIViewContentModeCenter;
+            self.imageView.backgroundColor = [UIColor blackColor];
+            if (_asset.assetType == WPMediaTypeImage) {
+                self.image = [WPMediaPickerResources imageNamed:@"camera" withExtension:@"png"];
+            } else if (_asset.assetType == WPMediaTypeVideo) {
+                self.image = [WPMediaPickerResources imageNamed:@"video" withExtension:@"png"];
+            }
             return;
         }
+        self.imageView.contentMode = UIViewContentModeScaleAspectFill;
```
frosty: I wonder if we should do this either in `prepareForReuse` or `setImage:animated:` instead?

Pod/Classes/WPMediaCollectionViewCell.m
```diff
@@ -193,8 +193,17 @@ - (void)fetchAssetImage
     requestKey = [_asset imageWithSize:requestSize completionHandler:^(UIImage *result, NSError *error) {
         if (error) {
```
frosty: Should we also check `result != nil`?

Pod/Classes/WPMediaCollectionViewCell.m
```diff
@@ -193,8 +193,17 @@ - (void)fetchAssetImage
...
             self.image = nil;
+            self.imageView.contentMode = UIViewContentModeCenter;
+            self.imageView.backgroundColor = [UIColor blackColor];
+            if (_asset.assetType == WPMediaTypeImage) {
+                self.image = [WPMediaPickerResources imageNamed:@"camera" withExtension:@"png"];
```
frosty: Should we use `setImage:` here?

frosty:
Sorry, just thought of two other things :)

SergioEstevao:

Pod/Classes/WPMediaCollectionViewCell.m
```diff
@@ -148,6 +150,44 @@ - (void)displayAudioAssetTypePlaceholder
...
+    NSTimeInterval duration = [self.asset duration];
+    [self setCaption:[self stringFromTimeInterval:duration]];
+}
+
+- (void)displayImageAssetTypePlaceholder
+{
+    self.placeholderStackView.hidden = NO;
+    self.imageView.hidden = YES;
+
+    if ([self.asset respondsToSelector:@selector(fileExtension)]) {
```
frosty: Perhaps we could extract this to some kind of `showAssetFileExtensionWithDefault:` method or something?

Pod/Classes/WPMediaCollectionViewCell.m
```diff
@@ -157,13 +197,13 @@ - (void)setAsset:(id<WPMediaAsset>)asset {
...
             [self fetchAssetImage];
-
             label = [NSString stringWithFormat:NSLocalizedString(@"Image, %@", @"Accessibility label for image thumbnails in the media collection view. The parameter is the creation date of the image."), formattedDate];
         break;
         case WPMediaTypeVideo:
+            [self displayVideoAssetTypePlaceholder];
             [self fetchAssetImage];
-
             label = [NSString stringWithFormat:NSLocalizedString(@"Video, %@", @"Accessibility label for video thumbnails in the media collection view. The parameter is the creation date of the video."), formattedDate];
             NSTimeInterval videoDuration = [asset duration];
```
frosty: You have this same code in `displayVideoAssetTypePlaceholder`, so I guess it's not necessary here now?

SergioEstevao:

frosty:
So I think the most recent changes don't behave as well as what we had before, on device. 

Scrolling doesn't seem as smooth, and I think it looks a little odd seeing a flash of the camera icon and "PHOTO" before the photo loads. Also, images are now displaying their filenames in the caption, which we don't want.

I think maybe we should either revert to the previous approach or see if we can address those issues.

frosty:
Performance is much improved with the latest changes. One remaining issue for me – video durations aren't displaying until I select a cell. The black translucent caption bar is visible, but not the label / time.

SergioEstevao:
@frosty ready for another round of review.

frosty:
Great! This fixes the issues I was seeing! :shipit:

PR #171, opened 2017-05-03
Refactor date formatting

SergioEstevao:
This PR just does some refactor around date/time formatting code to avoid code duplication and make things more consistent for our users.

To test:
 - Run the demo app
 - See if the time on videos shows correctly on thumbs cells and when previewing video full screen.


frosty:
PR looks great, nice cleanup of the code!

While testing, I found a small issue with the time formatting, which perhaps we could clean up at the same time?

img_7067

SergioEstevao:

Example/Tests/WPDateTimeHelpersTests.m
```diff
@@ -0,0 +1,35 @@
+#import <XCTest/XCTest.h>
+#import "WPDateTimeHelpers.h"
+
+@interface NSString ()
++ (NSString *)emojiCharacterFromCoreEmojiFilename:(NSString *)filename;
```
SergioEstevao: Bad copy paste...

SergioEstevao:
I will go for the formatter approach looks more consistent.

SergioEstevao:

PR #176, opened 2017-05-04
Improve data fetch performance

SergioEstevao:
This PR tries to further address performance issues when trying to display the Photo Library data.

It implements three things:

 - Make sure that the data update is made on a background thread.
 - Only fetch group data (count and poster) when needed.
 - Don't reload collection when refresh starts.

To test:
 - Open the demo app and see if the initial display and scrolling performance is good
 - Switch collection/groups/albums to see if all works correctly.
 - Capture an image using the capture header and check if all updates correctly.
 - Capture an image using the screenshot capture to see if all updates correctly


frosty:
Thanks for all the investigation on this, @SergioEstevao! I think we're good to :shipit:?

PR #177, opened 2017-05-05
Fix cases where asset date returns nil.

SergioEstevao:
Some assets can return nil dates and our formatters for user friendly dates where complaining this was an error and in the future will be an exception/crash.

To avoid this situation I added some checks for nil, an Unit Test and some nil handling on the project.

To test:
 - Run unit tests to check if no error occurs.
 - Double check on the main project if no exception when scrolling media libraries.

### DIFF
--- a/Pod/Classes/WPMediaCollectionViewCell.m
+++ b/Pod/Classes/WPMediaCollectionViewCell.m
@@ -122,7 +122,7 @@ static const CGFloat TimeForFadeAnimation = 0.3;
 - (void)configureAccessibility
 {
     NSString *accessibilityLabel = @"";
-    NSString *formattedDate = NSLocalizedString(@"Unknow creation date", @"Label to use when creation date from media asset is not know.");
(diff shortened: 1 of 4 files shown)


Pod/Classes/WPMediaCollectionViewCell.m
```diff
@@ -122,7 +122,11 @@ - (void)commonInit
 - (void)configureAccessibility
 {
     NSString *accessibilityLabel = @"";
-    NSString *formattedDate = [NSString stringWithFormat:@"%@ %@",[WPDateTimeHelpers userFriendlyStringDateFromDate:_asset.date], [WPDateTimeHelpers userFriendlyStringTimeFromDate:_asset.date]];
+    NSString *formattedDate = NSLocalizedString(@"Unknow creation date", @"Label to use when creation date from media asset is not know.");
```
frosty: Typo: Unknown

PR #182, opened 2017-06-01
Improve startup time of the picker.

### LINKED ISSUE #181: Improve startup time performance of the picker
At the moment the picker can take up to 2s to load. Let's try to make this quicker.

SergioEstevao:
Fixes #181 

To test:
 - Start demo app
 - Start the full screen picker
 - Check if performance improved.


frosty:
Let's go ahead and merge this, as it seems to fix up a couple of issues with initialization. It doesn't _drastically_ speed up startup for me – in some rough tests it was giving me 2 - 2.5 seconds instead of 3.5 seconds on develop. But everything helps!

PR #186, opened 2017-07-12
Move all picker options to separate class.

### LINKED ISSUE #185: Refactor the multiple options on WPMediaPickerViewController to a single struct/class
I purpose to refactor/move to an options object, ```WPMediaPickerOptions```, the following attributes :

- ```showMostRecentFirst```
- ```filter```
- ```allowMultipleSelection```
- ```cameraPreviewSize```
- ```preferFrontCamera```
- ```allowCaptureOfMedia```

This will allow to make it easier to add new attributes to it, and reflect them on the classes that use it  like `WPNavigationMediaPickerViewController` and `WPInputMediaPickerViewController`.

The options object could be set on the init methods, but also be changed with a setter method, that will make sure the picker get's updated correctly in a single place.

SergioEstevao:
Fixes #185 

To test:
 - Use the demo app and see if all is working correctly.

### DIFF
--- a/Pod/Classes/WPMediaPickerOptions.h
+++ b/Pod/Classes/WPMediaPickerOptions.h
@@ -1,20 +1,20 @@
 #import <Foundation/Foundation.h>
 #import "WPMediaCollectionDataSource.h"
 
-@interface WPMediaPickerOptions: NSObject
+@interface WPMediaPickerOptions: NSObject<NSCopying>
 
 /**
- If set the picker will show a cell that allows capture of new media, that can be used immediatelly
+ If YES the picker will show a cell that allows capture of new media, that can be used immediatelly
  */
 @property (nonatomic, assign) BOOL allowCaptureOfMedia;
 
 /**
- If the media picker allows media capturing, it will use the front camera by default when possible
+ If YES and the media picker allows media capturing, it will use the front camera by default when possible
  */
 @property (nonatomic, assign) BOOL preferFrontCamera;
 
 /**
- If set the picker will show the most recent items on the top left. If not set it will show on the bottom right. Either way it will always scroll to the most recent item when showing the picker.
+ If YES the picker will show the most recent items on the top left. If not set it will show on the bottom right. Either way it will always scroll to the most recent item when showing the picker.
  */
 @property (nonatomic, assign) BOOL showMostRecentFirst;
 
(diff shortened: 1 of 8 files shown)


Pod/Classes/WPMediaPickerOptions.h
```diff
@@ -0,0 +1,36 @@
+#import <Foundation/Foundation.h>
+#import "WPMediaCollectionDataSource.h"
+
+@interface WPMediaPickerOptions: NSObject
+
+/**
+ If set the picker will show a cell that allows capture of new media, that can be used immediatelly
+ */
+@property (nonatomic, assign) BOOL allowCaptureOfMedia;
```
frosty: I wonder if we should rename this to something like `showCaptureCell` or `showCameraPreview` or something.

frosty: For each of these BOOL properties, I'd say "If YES" not "if set", just because it's a bit more explicit. You could set it to `NO`, after all.

Pod/Classes/WPNavigationMediaPickerViewController.m
```diff
@@ -53,12 +49,8 @@ - (UIViewController *)childViewControllerForStatusBarStyle
 
 - (void)setupNavigationController
 {
-    WPMediaPickerViewController *vc = [[WPMediaPickerViewController alloc] init];
-    vc.allowCaptureOfMedia = self.allowCaptureOfMedia;
-    vc.preferFrontCamera = self.preferFrontCamera;
-    vc.showMostRecentFirst = self.showMostRecentFirst;
-    vc.filter = self.filter;
-    vc.allowMultipleSelection = self.allowMultipleSelection;
+    WPMediaPickerViewController *vc = [[WPMediaPickerViewController alloc] initWithOptions:self.options];
```
frosty: Much simpler 👍

Pod/Classes/WPMediaPickerOptions.h
SergioEstevao: Regarding the allowCaptureOfMedia, I'm not sure we should change it we are not only showing the CaptureCell or preview we are actually allowing users to capture new assets to the datasource.

SergioEstevao:
@frosty thanks for the review, I pushed some changes to address your comments please have a look.

frosty:
Looks good, happy to stick with the property name as-is for now :)

:shipit:!

PR #190, opened 2017-07-20
Break retain cycles.

### LINKED ISSUE #128: Crash in WPMediaCollectionViewController updateDataWithRemoved:inserted:changed:moved

SergioEstevao:
Fixes #128 

Found the following retain cycles:
 - Changing a group was making a retain cycle between the group picker and the data source
 - When refreshing data on the main collection view there was a retain cycle between the picker and the data source.

This retains cycles where keeping alive collection picker and data sources that then get notification messages out of place.

To test:
 - Open the demo app
 - Tap on Add
 - Capture photos, change groups, select photos
 - See that no crash happens.

### DIFF
--- a/Pod/Classes/WPMediaGroupPickerViewController.h
+++ b/Pod/Classes/WPMediaGroupPickerViewController.h
@@ -13,7 +13,7 @@
 /**
  The WPMediaCollectionDataSource that is being used to display the assets and groups. If not set the picker will create a new one.
  */
-@property (nonatomic, strong) id<WPMediaCollectionDataSource> dataSource;
+@property (nonatomic, weak) id<WPMediaCollectionDataSource> dataSource;
 
 @end
 
--- a/Pod/Classes/WPMediaPickerViewController.m
+++ b/Pod/Classes/WPMediaPickerViewController.m
@@ -91,7 +91,7 @@ static CGFloat SelectAnimationTime = 0.2;
     __weak __typeof__(self) weakSelf = self;
     self.changesObserver = [self.dataSource registerChangeObserverBlock:
                             ^(BOOL incrementalChanges, NSIndexSet *removed, NSIndexSet *inserted, NSIndexSet *changed, NSArray *moves) {
-                                if (incrementalChanges && !self.refreshGroupFirstTime) {
+                                if (incrementalChanges && !weakSelf.refreshGroupFirstTime) {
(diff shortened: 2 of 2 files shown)


SergioEstevao:
This can be further tested on the main WP-iOS app using this branch: issue/media_picker_crash_fix

frosty:
Thanks for picking this up, @SergioEstevao. I've tested this against `develop` (both here and the WPiOS branch) and I can't recreate a crash I could previously recreate on WPiOS.

However, I don't think I'm seeing much of a difference in terms of retained objects staying in memory? What exactly was hanging around before? What should I look out for being different with this branch?

One thing I did notice in WPiOS is that the order of items in the picker is reversed. I wonder if there's an issue with the new options object not working correctly.

SergioEstevao:
> However, I don't think I'm seeing much of a difference in terms of retained objects staying in memory? What exactly was hanging around before? What should I look out for being different with this branch?

I had a look and the difference was more noticeable on Aztec because we where keeping the reference to the data source on the AztecPostViewController. By keeping the reference to the datasource we where keeping references to the WPNavigationMediaPickerViewController and objects inside.

Another thing that was kept around was the group selector picker (WPMediaGroupPickerViewController), but again because we where keeping the data source around.
This one should be noticeable even on the demo app of the MediaPicker, every time we changed group/album one instance of those should stay around.

> One thing I did notice in WPiOS is that the order of items in the picker is reversed. I wonder if there's an issue with the new options object not working correctly.

That was a missing configuration step on the options object on the hybrid editor.

frosty:
Thanks for clarifying – I tested the changes more on Aztec and I definitely the difference in the memory graph now. Let's merge this!

:shipit:

PR #191, opened 2017-07-20
Improve options interface

SergioEstevao:
Refactor some of the code around options to make it more consistent across the different ViewControllers.

They now all have an init method that accepts the option object.
To make further changes you need to go directly to the mediaPicker object directly.

To test:
 - Run the demo app
 - Check if options are being applied correctly to the different pickers.

### DIFF
--- a/Pod/Classes/WPMediaPickerViewController.m
+++ b/Pod/Classes/WPMediaPickerViewController.m
@@ -108,12 +108,16 @@ static CGFloat SelectAnimationTime = 0.2;
 }
 
 - (void)setOptions:(WPMediaPickerOptions *)options {
+    WPMediaPickerOptions *originalOptions = _options;
     _options = [options copy];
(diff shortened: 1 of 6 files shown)


Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -108,12 +108,16 @@ - (void)viewDidLoad
 }
 
 - (void)setOptions:(WPMediaPickerOptions *)options {
+    WPMediaPickerOptions *originalOptions = _options;
     _options = [options copy];
+    BOOL refreshNeeded = (originalOptions.filter != options.filter) | (originalOptions.showMostRecentFirst != options.showMostRecentFirst) | (originalOptions.allowCaptureOfMedia != options.allowCaptureOfMedia);
```
frosty: We could implement `isEqual` on `WPMediaPickerOptions` and then just do `if (_options != options)`

SergioEstevao: Some of the options don't need to force a refresh, like allowsMultipleSelection.

On the main app on the MediaLibrary if I didn't implement it like this you will see a "flash", when switching from Edit Mode to View Mode.

frosty: Okay, let's leave it as-is :) :shipit:

PR #192, opened 2017-07-25
Improve selected assets

SergioEstevao:
This PR refactor the selectedAssets method to allow changed of the selectedAssets to come from the outside of the picker and/or to be configured before the picker is show.

To test:
 - Start the demo app
 - Start the input picker
 - Select some assets
 - Tap on + to start the full screen picker
 - See if the selected assets options where passed correctly
 - Tap and untap assets and see if they updated correctly

### DIFF
--- a/Example/WPMediaPicker/DemoViewController.m
+++ b/Example/WPMediaPicker/DemoViewController.m
@@ -237,7 +237,7 @@
     self.mediaPicker.dataSource = self.pickerDataSource;
     if (self.mediaInputViewController) {
         self.mediaPicker.mediaPicker.selectedAssets = self.mediaInputViewController.mediaPicker.selectedAssets;
-        self.mediaInputViewController.mediaPicker.selectedAssets = (NSArray<WPMediaAsset> *)@[];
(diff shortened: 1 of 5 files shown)


frosty:
Looking good, but found one bug:

* Open the input picker, select a couple of items
* Open the full screen picker, check items are still selected. Select a few extra items.
* Tap Cancel
* Tap + again to open the full screen picker
* It'll reshow with the original selection. I think it should either have no selection (probably the right option) or the updated selection.

SergioEstevao:
@frosty bug fixed, please have another look :)

Example/WPMediaPicker/DemoViewController.m
```diff
@@ -237,6 +237,7 @@ - (void)showPicker:(id) sender
     self.mediaPicker.dataSource = self.pickerDataSource;
     if (self.mediaInputViewController) {
         self.mediaPicker.mediaPicker.selectedAssets = self.mediaInputViewController.mediaPicker.selectedAssets;
+        self.mediaInputViewController.mediaPicker.selectedAssets = (NSArray<WPMediaAsset> *)@[];
```
frosty: Nitpicking really, but I think `[NSArray<WPMediaAsset> new]` would be neater?

PR #197, opened 2017-07-31
Remove camera cell from the inline picker

bummytime:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/7628

This PR removes the camera cell from the inline media picker by overriding the `allowCaptureOfMedia` option upon init `WPInputMediaPickerViewController`.

napkin 45 07-31-17 3 40 38 pm

You can also test this change in WPiOS via [this branch](https://github.com/wordpress-mobile/WordPress-iOS/tree/issue/7628-remove-camera-cell-from-inline-picker). 

wp_1

**To test:**
 Run the demo on a few devices (ideally an iPad and iPhone):
1. Tap the quick select assets cell in the table. Make sure the camera cell is not visible. 
2. Tap the plus icon in the upper right corner, verify the camera cell is visible.

@frosty Could you give this a looksie? Thank you!

### DIFF
--- a/Example/WPMediaPicker/DemoViewController.m
+++ b/Example/WPMediaPicker/DemoViewController.m
@@ -275,7 +275,6 @@
 {
     if (textField == self.quickInputTextField) {
         [self setupMediaKeyboardForInputField];
-        self.mediaInputViewController.mediaPicker.options = [self selectedOptions];        
         [self.mediaInputViewController.mediaPicker resetState:NO];
     }
     return YES;
--- a/Example/WPMediaPicker/DemoViewController.m
+++ b/Example/WPMediaPicker/DemoViewController.m
@@ -275,6 +275,7 @@
 {
     if (textField == self.quickInputTextField) {
         [self setupMediaKeyboardForInputField];
+        self.mediaInputViewController.options = [self selectedOptions];
         [self.mediaInputViewController.mediaPicker resetState:NO];
     }
     return YES;
--- a/Pod/Classes/WPInputMediaPickerViewController.h
+++ b/Pod/Classes/WPInputMediaPickerViewController.h
@@ -36,6 +36,12 @@ The delegate for the WPMediaPickerViewController events
 @property (nonatomic, readonly)  WPMediaPickerViewController * _Nonnull mediaPicker;
 
 /**
+ Options passed to the internal media picker instance. Options should be set via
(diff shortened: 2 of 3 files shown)


frosty:
Thanks for picking this up so quickly, @bummytime!

Looks good in WPiOS, but in the Media Picker demo there seems to be an issue where the options aren't applied correctly to the picker (I think this is from the change you made in `textFieldShouldBeginEditing:`). If you open the inline picker in the demo app, you'll notice that the scroll direction is backwards (it begins scrolled to the right) – this is because the `showMostRecentFirst` option isn't set right.

I wonder if we could add an `options` property to the input picker view, and use that to customise options as they're passed in. Something like this? https://gist.github.com/frosty/f039793936682651e533272ae12534ba

Let me know if none of that makes sense!

SergioEstevao:
Isn't there an options on the Init method? That allows configuration of this?

Enviado do meu iPhone

No dia 1 de ago de 2017, às 11:09, James Frost <notifications@github.com> escreveu:

bummytime:
@frosty Thanks for the quick review. Yeah, I totally missed that — I removed that line in `textFieldShouldBeginEditing:` because it was stomping on the options override being set in `init`. Your diff looks workable and I will update this PR. 🙇

bummytime:
@frosty, changes made. Would you mind giving this PR another look? Thanks!

bummytime:
Thanks @frosty!

frosty:
@sergioestevao I suggested to @bummytime about how to implement this. The idea is that the live cell in the inline picker will be replaced by a camera button in the bottom toolbar (if the `allowsCaptureOfMedia` option is true). 

If we want to allow either to be used, I guess we'd need to add an extra option - but I'm not sure it's worth the effort if we won't be using that option ourselves?

SergioEstevao:
I will prefer to have an extra separated option for the toolbar, I think we
are loosing a possible configuration that users of the library would like.
I think the extra work to have this options will be very small.

bummytime:
@SergioEstevao — thanks for your feedback and [sample PR on WPiOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/7643). @frosty and I chatted about this a bit and thought you made a good point. Soooo, I will revert this PR and will use the existing option to turn off the media capture cell. 😄

PR #199, opened 2017-08-04
Use long press to preview in the inline picker

bummytime:
This PR overrides the `WPInputMediaPickerViewController`'s trait collection to disable 3D Touch so preview can work properly when the picker is used in an `inputView`. 

In addition, when displaying the preview controller itself, we default to `self.viewControllerToUseToPresent.navigationController` instead of `self.navigationController` because `self.navigationController` will be nil when the picker as an input view.

test

Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/7632

To test:

1. Review previewing in the demo app's quick select assets field. Verify long pressing and preview display works on all devices...with or without 3D Touch.
2. Install WPiOS from the the [picker's test branch](https://github.com/wordpress-mobile/WordPress-iOS/tree/issue/inline-media-picker-updates). 
3. Using a debug build, test the inline picker. Verify long pressing and preview display works on all devices...with or without 3D Touch.

@frosty Could I bother you for a review? Thanks!!

### DIFF
--- a/Pod/Classes/WPInputMediaPickerViewController.m
+++ b/Pod/Classes/WPInputMediaPickerViewController.m
@@ -49,8 +49,9 @@
     self.privateDataSource = [[WPPHAssetDataSource alloc] init];    
     self.mediaPicker.dataSource = self.privateDataSource;
 
-
     [self addChildViewController:self.mediaPicker];
+    [self overridePickerTraits];
+    
     self.mediaPicker.view.frame = self.view.bounds;
     self.mediaPicker.view.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
     [self.view addSubview:self.mediaPicker.view];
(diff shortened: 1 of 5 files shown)


Pod/Classes/WPInputMediaPickerViewController.m
```diff
@@ -82,6 +83,20 @@ - (void)viewDidLayoutSubviews {
...
+- (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection
+{
+    [super traitCollectionDidChange:previousTraitCollection];
+    [self overridePickerTraits];
+}
+
+- (void)overridePickerTraits
+{
+    // Due to an inputView being displayed in its own window, the force touch peek transition
+    // doesn't display correctly. Because of this, we'll disable
```
frosty: I think the end of this sentence got cut off?

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -818,6 +815,39 @@ - (UIViewController *)defaultPreviewViewControllerForAsset:(id <WPMediaAsset>)as
     return fullScreenImageVC;
 }
 
+- (void)displayPreviewController:(UIViewController *)viewController {
+    if (viewController) {
+        // viewControllerToUseToPresent defaults to self but could be set to nil. Reset to self if needed.
+        if (!self.viewControllerToUseToPresent) {
```
frosty: We could put this on a getter for `viewControllerToUseToPresent`, perhaps?

```
- (UIViewController *)viewControllerToUseToPresent
{
    if (!_viewControllerToUseToPresent) {
        _viewControllerToUseToPresent = self;
    }

    return _viewControllerToUseToPresent;
}
```

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -818,6 +815,39 @@ - (UIViewController *)defaultPreviewViewControllerForAsset:(id <WPMediaAsset>)as
...
+                                                                                                            target:self
+                                                                                                            action:@selector(dismissPreviewController)];
+            [self.viewControllerToUseToPresent presentViewController:navController animated:YES completion:nil];
+        }
+    }
+}
+
+- (void)dismissPreviewController {
+    // viewControllerToUseToPresent defaults to self but could be set to nil. Reset to self if needed.
+    if (!self.viewControllerToUseToPresent) {
```
frosty: This can be removed if we move code to the getter, as above.

PR #200, opened 2017-08-07
Updates inline picker thumbnail size and scroll direction

### LINKED ISSUE #196: Add option for media input picker to show assets in vertical mode.
At the moment the input media picker only allows to show assets in an horizontal scrolling mode, it will be good to also allow to show in vertical scrolling mode.

bummytime:
This PR:

1. Changes the scroll direction of the inline picker to vertical.
2. Adjusts the number of image cells per row based on the width of the frame which contains the input picker. Devices: 
  * SE = 3-wide portrait / 5-wide landscape
  * 7/7+ = 4-wide portrait / 6-wide landscape
  * iPad @ 768 = 7-wide
  * iPad @ 1024 = 9-wide

napkin 47 08-07-17 4 45 35 pm
napkin 47 08-07-17 4 46 40 pm

Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/7630
Fixes #196 

**To test:** 
1. On a variety of devices run the picker demo app, verify the picker grid scrolls vertically and the correct number of images are displayed in each orientation.
2. Download and run [the test branch for WPiOS](https://github.com/wordpress-mobile/WordPress-iOS/tree/issue/inline-media-picker-updates). Verify the picker grid scrolls vertically and the correct number of images are displayed in each orientation.

@kwonye would you be able to give this a quick 👀 ?

cc/ @iamthomasbishop

### DIFF
--- a/Pod/Classes/WPInputMediaPickerViewController.m
+++ b/Pod/Classes/WPInputMediaPickerViewController.m
@@ -1,6 +1,13 @@
 #import "WPInputMediaPickerViewController.h"
 #import "WPPHAssetDataSource.h"
 
+static CGFloat const IPhoneSELandscapeWidth = 568.0f;
+static CGFloat const IPhone7PortraitWidth = 375.0f;
+static CGFloat const IPhone7LandscapeWidth = 667.0f;
+static CGFloat const IPadPortraitWidth = 768.0f;
+static CGFloat const IPadLandscapeWidth = 1024.0f;
+static CGFloat const IPadPro12LandscapeWidth = 1366.0f;
+
 @interface WPInputMediaPickerViewController()
 
 @property (nonatomic, strong) WPMediaPickerViewController *mediaPicker;
@@ -55,9 +62,6 @@
     self.mediaPicker.view.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
     [self.view addSubview:self.mediaPicker.view];
     [self.mediaPicker didMoveToParentViewController:self];
-    self.mediaPicker.collectionView.bounces = NO;
-    self.mediaPicker.collectionView.alwaysBounceHorizontal = NO;
-    self.mediaPicker.collectionView.alwaysBounceVertical = NO;
 
     self.mediaToolbar = [[UIToolbar alloc] initWithFrame:CGRectMake(0, 0, self.view.frame.size.width, 44)];
     self.mediaToolbar.items = @[
@@ -101,10 +105,14 @@
 
         layout.scrollDirection = UICollectionViewScrollDirectionVertical;
         layout.sectionInset = UIEdgeInsetsMake(2, 0, 0, 0);
+        self.mediaPicker.collectionView.alwaysBounceHorizontal = NO;
+        self.mediaPicker.collectionView.alwaysBounceVertical = YES;
     } else {
         photoSize = floorf((self.view.frame.size.height - photoSpacing) / 2.0);
         layout.scrollDirection = UICollectionViewScrollDirectionHorizontal;
         layout.sectionInset = UIEdgeInsetsMake(0, 5, 0, 5);
+        self.mediaPicker.collectionView.alwaysBounceHorizontal = YES;
+        self.mediaPicker.collectionView.alwaysBounceVertical = NO;
     }
 
     layout.itemSize = CGSizeMake(photoSize, photoSize);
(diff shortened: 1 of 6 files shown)


Pod/Classes/WPInputMediaPickerViewController.m
```diff
@@ -68,19 +67,69 @@ - (void)setupMediaPickerViewController {
...
+ Returns the maximum number of photos to be used in a picker row given the provided frame width.
+ 
+ @param frameWidth Width size of the frame containing the picker
+
+ @return The number of photo cells to be used in a row. Defaults to 3.
+ */
+- (NSUInteger)numberOfPhotosPerRow:(CGFloat)frameWidth {
+    NSUInteger numberOfPhotos = 3;
+
+    if (frameWidth >= 375 && frameWidth < 500) {
```
SergioEstevao: Could we have this based on size traits? Really don't like have this based on fixed numbers...

Pod/Classes/WPInputMediaPickerViewController.m
```diff
@@ -68,19 +67,69 @@ - (void)setupMediaPickerViewController {
...
+    }
+    
+    // Init and configure collection view layout
     UICollectionViewFlowLayout *layout = [[UICollectionViewFlowLayout alloc] init];
-    layout.scrollDirection = UICollectionViewScrollDirectionHorizontal;
-    layout.itemSize = CGSizeMake(size, size);
-    layout.minimumLineSpacing = spacing;
-    layout.minimumInteritemSpacing = spacing;
-    layout.sectionInset = UIEdgeInsetsMake(0, 5, 0, 10);
+    layout.scrollDirection = UICollectionViewScrollDirectionVertical;
```
SergioEstevao: Could we have an option on the VC to allow horizontal or vertical scrolling? This will allow more flexibility for users of the library.

Pod/Classes/WPInputMediaPickerViewController.m
```diff
@@ -35,7 +35,6 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
     return self;
 }
 
-
```
kwonye: I know this is tiny, but thanks for fixing this 😊

Pod/Classes/WPInputMediaPickerViewController.m
kwonye: Due to the amount of users we have of this cocoapods, I'm inclined to agree (Cocoapods says it's in 86 apps)

Pod/Classes/WPInputMediaPickerViewController.m
kwonye: If this is possible, that'd be great. Unfortunately from what I found online, size traits don't give us the level of details we require for this. @SergioEstevao do you know if that's true?

Pod/Classes/WPInputMediaPickerViewController.m
bummytime: Great idea @SergioEstevao! WIll do that.

Pod/Classes/WPInputMediaPickerViewController.m
bummytime: @SergioEstevao and @kwonye I felt the same way as I wrote this. Size classes would work for some of this, however, there are circumstances where they would not work. An example would be the iPhone SE and iPhone 7+ in portrait — both are considered "compact width". Per the designs, we needed a little more granular control over the precise width of the frame (e.g. 3-wide for the SE and 4-wide for the 7/7+), hence the frame sizes here. 

I can look into this a little more today.

kwonye: Cool, thanks @bummytime. Whatever you choose I trust your judgement.

bummytime:
@kwonye @SergioEstevao I made vertical scrolling in the picker optional now. By default, it will continue to scroll horizontally (and size the collection view items the "old" way) unless you explicitly set `scrollVertically` to `YES`. In that case, we will also use the new item sizing logic.

In addition, I added an option to the demo app to adjust scrolling on the quick input view:

napkin 47 08-09-17 9 28 45 am

bummytime:
@kwonye Filed that multitasking issue is here: #201. We can take care of that in a separate PR.  I was not able to find a better solution for the `numberOfPhotosPerRow` method. For now, it works and will only be used when vertical scrolling is turned on (and that is now `NO` by default.)

Ready for another 👀! 

😄

Pod/Classes/WPInputMediaPickerViewController.m
```diff
@@ -68,19 +69,75 @@ - (void)setupMediaPickerViewController {
...
+ to be used in a picker row.
+ 
+ @param frameWidth Width of the frame containing the picker
+
+ @return The number of photo cells to be used in a row. Defaults to 3.
+ */
+- (NSUInteger)numberOfPhotosPerRow:(CGFloat)frameWidth {
+    NSUInteger numberOfPhotos = 3;
+
+    if (frameWidth >= 375 && frameWidth < 500) {
```
kwonye: The only request I have since this is the best way to do this is to convert these numbers to constants please so that they're not just magic numbers to someone jumping in here.

PR #205, opened 2017-08-16
Update media cell selection position

### LINKED ISSUE #204: Implement new selection indicator new appearance
The selection indicator should look like the following mockups:

screen shot 2017-08-15 at 21 39 49

screen shot 2017-08-15 at 21 39 54

Unselected dot

- 21x21pt
- Background: rgba(198, 198, 198, 0.3)
- Border (inset): 1pt #FFFFFF
- Blur: 10px
- Placement: 10pt off bottom-left corner

Selected Dot

- 21x21pt
- Background: #00AADC
- Box-shadow:  0 0 8pt rgba(46, 68, 83, 0.4)
- No border
- Placement: 10pt off bottom-left corner
- Label font: SF UI Text Medium @ 13pt - centered in dot

SergioEstevao:
Fixes #204 

This PR updates the appearance of the cell image selector/position to the new design.

It now looks like this:
img_1521

To test:
 - Start the demo app
 - On the keyboard picker and on the full screen mode
 - Select and unselect cells and see if all works correctly

@iamthomasbishop: Do you think the 10pt distance from the edge looks ok? Do you think we need the shadow on the position button? what about to always keep the white border?

@frosty, performance wise to you the cornerRadius option is making things slow while scrolling? I can give a try to shapeLayer if you think it's too slow.

### DIFF
--- a/Pod/Classes/WPMediaCollectionViewCell.m
+++ b/Pod/Classes/WPMediaCollectionViewCell.m
@@ -95,7 +95,7 @@ static const CGFloat TimeForFadeAnimation = 0.3;
     self.selectedBackgroundView = _selectionFrame;
 
     CGFloat labelTextSize = 12.0;
-    CGFloat labelHeight = 30.0;    
+    CGFloat labelHeight = 30.0;
     CGColorRef topGradientColor = [[UIColor colorWithWhite:0 alpha:0] CGColor];
     CGColorRef bottomGradientColor = [[UIColor colorWithWhite:0 alpha:0.5] CGColor];
 
@@ -319,6 +319,9 @@ static const CGFloat TimeForFadeAnimation = 0.3;
 
 - (void)setSelected:(BOOL)selected
 {
+    if (selected == self.isSelected) {
+        return;
+    }
     [super setSelected:selected];
     [self updatePositionLabelToSelectedState:self.isSelected];
 }
--- a/Pod/Classes/WPMediaCollectionViewCell.m
+++ b/Pod/Classes/WPMediaCollectionViewCell.m
@@ -90,7 +90,9 @@ static const CGFloat TimeForFadeAnimation = 0.3;
     _positionLabel.clipsToBounds = YES;
     _positionLabel.textColor = [UIColor whiteColor];
     _positionLabel.textAlignment = NSTextAlignmentCenter;
-    _positionLabel.font = [UIFont systemFontOfSize:13];
+    _positionLabel.font = [UIFont systemFontOfSize:13 weight:UIFontWeightMedium];
+    _positionLabel.adjustsFontSizeToFitWidth = YES;
+    _positionLabel.minimumScaleFactor = 0.5;
 
     _positionLabelShadowView = [[UIView alloc] initWithFrame:_positionLabel.frame];
     _positionLabelShadowView.autoresizingMask = _positionLabel.autoresizingMask;
--- a/Pod/Classes/WPMediaCollectionViewCell.m
+++ b/Pod/Classes/WPMediaCollectionViewCell.m
@@ -4,6 +4,8 @@
 
 static const NSTimeInterval ThresholdForAnimation = 0.03;
 static const CGFloat TimeForFadeAnimation = 0.3;
+static const CGFloat LabelSmallFontSize = 9;
+static const CGFloat LabelRegularFontSize = 13;
 
 @interface WPMediaCollectionViewCell ()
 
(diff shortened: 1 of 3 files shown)


Pod/Classes/WPMediaCollectionViewCell.m
```diff
@@ -314,22 +320,31 @@ - (void)setPlaceholderTintColor:(UIColor *)placeholderTintColor
...
+    if (selected) {
+        _positionLabel.backgroundColor = [self tintColor];
+        _positionLabel.layer.borderColor = [self tintColor].CGColor;
+        _positionLabel.layer.shadowColor = [UIColor colorWithWhite:0 alpha:1].CGColor;
+        _positionLabel.layer.shadowRadius = 10.0;
+        _positionLabel.layer.shadowOffset = CGSizeMake(10, 10);
     } else {
-        
+        _positionLabel.text = @"";
+        _positionLabel.backgroundColor = [[UIColor lightGrayColor] colorWithAlphaComponent:0.7];
```
SergioEstevao: Good catch, I think I'm going to make this color configurable trough an appearance method to make it easier to configure for other users of the library.

Pod/Classes/WPMediaCollectionViewCell.m
```diff
@@ -314,22 +320,31 @@ - (void)setPlaceholderTintColor:(UIColor *)placeholderTintColor
...
 
+    [self updatePositionLabelToSelectedState:self.isSelected];
+}
+
+- (void)updatePositionLabelToSelectedState:(BOOL)selected
+{
+    if (selected) {
+        _positionLabel.backgroundColor = [self tintColor];
+        _positionLabel.layer.borderColor = [self tintColor].CGColor;
+        _positionLabel.layer.shadowColor = [UIColor colorWithWhite:0 alpha:1].CGColor;
```
SergioEstevao: I need to remove that shadowColor just realize that shadow is applied for the text on the label , and not for the border/circle.

Pod/Classes/WPMediaCollectionViewCell.m
```diff
@@ -314,22 +320,31 @@ - (void)setPlaceholderTintColor:(UIColor *)placeholderTintColor
 - (void)setSelected:(BOOL)selected
 {
     [super setSelected:selected];
-    if (self.isSelected) {
-    } else {
-        self.positionLabel.hidden = YES;
-    }
+    [self updatePositionLabelToSelectedState:self.isSelected];
```
SergioEstevao: Good catch.

SergioEstevao:
@frosty ready for another round.

SergioEstevao:
@iamthomasbishop updated screenshot with shadows on selection.

screen shot 2017-08-16 at 15 52 23

iamthomasbishop:
@SergioEstevao 
> Do you think the 10pt distance from the edge looks ok? Do you think we need the shadow on the position button? what about to always keep the white border?

I think 10pt from each edge looks good. I can't imagine there are any cases of videos having more than 5 digits, so the space should be sufficient.

Noticed you updated the shadow on selection dot 👍

iamthomasbishop:
@frosty 
> Also, whilst it's an unlikely situation I think we should probably do something about the case when the number of selected items is longer than 2 digits. Should let the label grow?

Good call. The label's font size should probably shrink in those cases to fit 3 digits, even if the fit is tight. The font size here can be 10pt. Example:

screenshot 2017-08-16 10 03 46

@SergioEstevao Another thing I noticed. Is the label set in Regular weight or is just my eyes playing tricks on me? It should be Medium.

SergioEstevao:
@iamthomasbishop and @frosty I setup the position to label to auto shrink it looks better but because of the circle mask applied to the label I see some numbers being cropped.

screen shot 2017-08-16 at 21 24 58

iamthomasbishop:
@SergioEstevao 
>  I setup the position to label to auto shrink it looks better but because of the circle mask applied to the label I see some numbers being cropped.

Looking good! Maybe we could come down 1pt more, to 9pt? I know that's really small, but I think this is appropriate for such an edge case (user selecting 100+ media items at a time).

Pod/Classes/WPMediaCollectionViewCell.m
```diff
@@ -308,6 +308,8 @@ - (void)setPosition:(NSInteger)position
 {
     _position = position;
     if (position != NSNotFound) {
+        CGFloat fontSize = position < 100 ? 13.0 : 9.0;
+        _positionLabel.font = [UIFont systemFontOfSize:fontSize weight:UIFontWeightMedium];
```
frosty: Could we pull the font sizes and weight out into constants so we only need to change them in one place? (we're setting 13 and medium in 2 different places)

I blame @jleandroperez for my pickiness about these things, sorry :)

SergioEstevao: @frosty Done! Just don't ask me to right align variable values :)

frosty: Haha! Awesome, thank you!

PR #207, opened 2017-08-17
Redesign group picker

### LINKED ISSUE #206: Implement group/album selector using full screen
Instead of showing a pop up for album selection inside the collection media picker, the album selector will now show full screen and will be the first thing show to the user.

Mockup:
screen shot 2017-08-17 at 16 54 54

SergioEstevao:
Fixes #206 

Implements new UI/UX of group picker.

To test:
 - Start the demo app
 - Open the full screen picker 
 - Check if album selector shows up
 - Select album
 - Select items
 - Check if Select option shows top left
 - Deselect Items
 - Check if Select option shows top left, and if selection zero it disappear
 - Press Select and see if assets are selected
 - Start picker again and press cancel.

@iamthomasbishop, @frosty  some notes:

 - Do you want me to implement the background images on the albums? I can do this by repeating the group image on smaller sizes on the back of the front image. I will use the same image to avoid extra downloads or export of media
- I added the possibility to configure the message for selection, by default it will be Select XX, but it can be configured to Insert, or another message that makes sense for the current use of the picker.


iamthomasbishop:
@SergioEstevao 
> Do you want me to implement the background images on the albums? I can do this by repeating the group image on smaller sizes on the back of the front image. I will use the same image to avoid extra downloads or export of media

Yea, go ahead with this :) Sounds good.

> I added the possibility to configure the message for selection, by default it will be Select XX, but it can be configured to Insert, or another message that makes sense for the current use of the picker.

Probably should be `Insert X`, as that's more explicit to what the action is.

frosty:
@SergioEstevao For the thumbnail stack, I wonder if you can use `fetchKeyAssetsInAssetCollection:options:` on `PHAsset` with a `fetchLimit` of e.g. 3 to quickly fetch the first 3 assets of a collection. Not tried it, just a guess!

SergioEstevao:
In the editor context insert is more logic, but remember the picker pod to is to be generic so Select makes more sense has the default. In the demo app I configured to be insert.

Enviado do meu iPhone

No dia 17 de ago de 2017, às 17:52, Thomas Bishop <notifications@github.com> escreveu:

SergioEstevao:
@frosty I can try, on device álbuns it should be quick, but it can get problematic on remote albums like the WordPress Library.

Enviado do meu iPhone

No dia 17 de ago de 2017, às 17:52, James Frost <notifications@github.com> escreveu:

frosty:
One more thing! The background for the album selector (and I guess the picker itself, actually) should be white and not blue :) is this configurable for WPiOS, or do we need to do it here?

frosty:
Essentially, for the thumbnail size / placement I think we should be trying to match the system UI as closely as possible (just, probably without the stacks to begin with):

screen shot 2017-08-18 at 10 52 28

SergioEstevao:
@frosty 

> Looking at the designs, I think we can make the album thumbnails larger – they look to be around 55pt here, and 70 or 71 in the mockup pngs. Also, I think the margins between the thumbnail and the left and right sides can be reduced.

I fixed this to follow the mockups.

> It might just be a demo app issue, but there's no permissions popup when you open the fullscreen picker and the app doesn't have permissions. The inline picker does show a popup.

I added the check for permission on the group picker.

> When I view the albums screen, it typically takes a good 5 seconds for anything at all to appear. I had a few times where it was much faster, but it's generally super slow with no indication that anything is happening. If nothing else, we should show a loading indicator, but I think we really need to work on improving the performance here before we ship this. I'm sure we must be able to get something to display almost instantly, even if just the album names and then we load in the counts / thumbnails. I'd really like us to ditch the loading indicators for the device photos albums, as I feel like they shouldn't be necessary – everything should be fast and always in sync anyway? Happy for us to address the loading time in a separate PR, before we ship the next release of the media picker.

Has we discussed we will address this on a separate PR.

>I'm seeing an issue when I select an album, tap Back, then select another album – I see a brief flash of the previous album as the new one is sliding into view. If it's hard to see by eye, you can record the navigation in Quicktime, and step through the frames with arrow keys. Most of the time it's a super brief flash, but I had at least one occurrence where it was there for a good second or so. Seems worse when switching between my top albums that have lots of images in (All Photos, Recently Added, Favourites, etc).

Fixed.

Pod/Classes/WPMediaGroupPickerViewController.m
```diff
@@ -52,7 +51,52 @@ - (void)viewDidLoad
...
     [self.tableView reloadData];
+    NSString *title = NSLocalizedString(@"Media Library", @"Title for alert when a generic error happened when loading media");
+    NSString *message = NSLocalizedString(@"There was a problem when trying to access your media. Please try again later.",  @"Explaining to the user there was an generic error accesing media.");
+    NSString *cancelText = NSLocalizedString(@"OK", "");
+    NSString *otherButtonTitle = nil;
+    if (error.domain == WPMediaPickerErrorDomain &&
+        error.code == WPMediaErrorCodePermissionsFailed) {
+        otherButtonTitle = NSLocalizedString(@"Open Settings", @"Go to the settings app");
+        title = NSLocalizedString(@"Media Library", @"Title for alert when access to the media library is not granted by the user");
+        message = NSLocalizedString(@"This app needs permission to access your device media library in order to add photos and/or video to your posts. Please change the privacy settings if you wish to allow this.",
```
frosty: I know we already use this text in the library, but we might want to check with @iamthomasbishop if it needs any adjustments – @iamthomasbishop, is the title and message text here okay for the 'you don't have permission' alert?

SergioEstevao:
@iamthomasbishop here is a screenshot of the current version.

img_1539

At the moment we are omitting the stack view on the group image. This is because of some issue with fetching multiple images to make that view.

Another question we have is on the final version on the app should the background of the view be gray? like the one in the asset view? or should be white on both?

iamthomasbishop:
@SergioEstevao 

I agree w/ James above:
> Essentially, for the thumbnail size / placement I think we should be trying to match the system UI as closely as possible

If we can match the stacked UI that's typical, that'd be preferred. If not possible, I can come up with a "fake" stacked UI where the underlying thumbs are just grayed out squares. But strongly prefer the typical stacked UI.

Also, fwiw in terms of thumb size, I'm designing on an iPhone Plus size canvas so I'm not sure if that affects the thumb size or if it stays as you go down in device size.

> on the final version on the app should the background of the view be gray? like the one in the asset view? or should be white on both?

Assuming by "asset view" you mean from the thumbnail picker - both should be white background. Example (photos.app)

img_0755

SergioEstevao:
@iamthomasbishop but on the WPiOS app will we do white background and grey navigation bar? this will look a part from the rest of the app. Remember this picker is not only used on the editor, it's used for featured image, gravatar icon, site icon picker.

frosty:
I think white background for the table and collection views, but stick with the Wordpress blue navigation bar?

iamthomasbishop:
@SergioEstevao 
> on the WPiOS app will we do white background and grey navigation bar?

sorry, I meant white for the body/table of the view - keep the WP Blue on the navigation bar.

SergioEstevao:
@iamthomasbishop ok, I will merge this PR now, and then do a separate one for the stack effect.

PR #208, opened 2017-08-17
Expose `showCapture` functionality to media picker and input picker

frosty:
This PR exposes the `showCapture` method in the interface of media picker and the input picker, so we can present the capture UI directly from the new inline picker format bar in WPiOS.

Not much to check here!

Needs review: @SergioEstevao


frosty:
Refs https://github.com/wordpress-mobile/WordPress-iOS/issues/7703

PR #209, opened 2017-08-18
Performance improvements on groups

SergioEstevao:
This PR addresses some performance issues on the group selector.

To test:
 - Open the full screen picker.
 - Check the performance when opening, and when opening albums.


frosty:
This is working really fast now against my media library! Thanks so much for diving into this, @SergioEstevao! Just need a background colour on the poster cells and I think we're there!

PR #212, opened 2017-08-24
Issue/fix header size on start

SergioEstevao:
When opening the inputPicker for the first time on the demo and the live preview cell was active, the cell got the wrong size. After some testing I found out that we had some code on the refreshing of the picker that was causing the cell to get the wrong size.

To test:
 - Start the demo app.
 - Tap on the text field.
 - See if the live cell has the correct size.

Note: I think this didn't happen before the performance improvements because it took so much time to display the album the header cell got the correct size meanwhile.


bummytime:
Nice work @SergioEstevao! :shipit:

PR #215, opened 2017-08-24
If the view is not loaded yet we don’t need to force a refresh.

SergioEstevao:
This PR implement a super small fix to avoid a double refresh when setting the group the first time.

To test:
 - Start the demo app
 - Open the full screen picker
 - Check if opening albums work ok.


bummytime:
Everything works great @SergioEstevao! :shipit:

PR #222, opened 2017-08-29
Improve comment for Select action.

SergioEstevao:
Improve localisation comment for easier translation.

### DIFF
--- a/Pod/Classes/WPNavigationMediaPickerViewController.m
+++ b/Pod/Classes/WPNavigationMediaPickerViewController.m
@@ -253,7 +253,7 @@ static NSString *const ArrowDown = @"\u25be";
 - (NSString *)selectionActionValue {
     NSString *actionString = self.selectionActionTitle;
     if (actionString == nil) {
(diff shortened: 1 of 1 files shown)


Pod/Classes/WPNavigationMediaPickerViewController.m
```diff
@@ -253,7 +253,7 @@ - (void)updateSelectionAction {
 - (NSString *)selectionActionValue {
     NSString *actionString = self.selectionActionTitle;
     if (actionString == nil) {
-        actionString = NSLocalizedString(@"Select %@", @"");
+        actionString = NSLocalizedString(@"Select %@", @"Action for Navigation bar to indicate selection of media. The argument in the string is a numeral number that indicates the number of elements selected.");
```
frosty: You can probably shorten the last bit of this to "The argument in the string represents the number of elements selected."

SergioEstevao: I wanted to say it's digits like this: 1 and not like this: one, two.

frosty: Ah. Perhaps we can say "The argument in the string represents the number of elements (as numeric digits) selected."

PR #228, opened 2017-08-31
Update selection when background changes to assets is made.

### LINKED ISSUE #225: When changes are done to the photos that effect selection, refresh was not done.
### Expected behavior

If some elements are selected, but you are on an album that doesn't reflect the selection if a change is done on the background ( delete the selected photos), selection info ( Insert X) should be updated.

### Actual behavior

Selection is not updated

### Steps to reproduce the behavior

- Open the demo app
- Go to camera roll and select two photos
- Switch to another album where selected photos are not part of it
- Switch to the Photos app
- Delete the selected photos
- Switch back to demo, see that selection is not change

SergioEstevao:
Fixes #225 

To test:
 - see ticket description above


frosty:
Thanks for catching this!

PR #231, opened 2017-09-01
Extract capture presentation

frosty:
This PR extracts the functionality for displaying the image capture interface from the media picker itself. This makes it easier to present it by itself from other places – in particular, from a new option we'll be adding to the hybrid editor in WPiOS.

To test:

* Build and run the demo app and check that image capture still works as expected.


frosty:
Thanks!

PR #232, opened 2017-09-04
Filter albums according to number of elements of filter type.

### LINKED ISSUE #230: Hide empty albums
Due to some changes we made to speed up the appearance of the picker, we're currently showing empty albums in the albums list:

fullsizerender 3

We should find a way to hide these empty albums.

SergioEstevao:
Fixes #230 

To test:
 - Open the demo app
 - Open the full screen picker
 - Check performance

@frosty do you mind to test this on your device? This is just to give it a first shot if it doesn't work feel free to test other options.


frosty:
Thanks for trying this, @SergioEstevao! Unfortunately, it seems to put us back in the 4-or-5 seconds to show any collections situation :( I guess it's because we're doing a fetch of the assets on each collection.

PR #233, opened 2017-09-06
Improve swift nullability annotations.

SergioEstevao:
To test:
 - Check if the app compiles properly without warning or errors
 - Run unit tests.

### DIFF
--- a/Pod/Classes/WPMediaCollectionDataSource.h
+++ b/Pod/Classes/WPMediaCollectionDataSource.h
@@ -8,7 +8,7 @@ typedef NS_OPTIONS(NSInteger, WPMediaType){
     WPMediaTypeAll= 0XFF
 };
 
-static NSString * const WPMediaPickerErrorDomain = @"WPMediaPickerErrorDomain";
+static NSString * _Nonnull const WPMediaPickerErrorDomain = @"WPMediaPickerErrorDomain";
 
 typedef NS_ENUM(NSInteger, WPMediaPickerErrorCode){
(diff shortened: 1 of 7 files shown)


Pod/Classes/WPMediaCollectionDataSource.h
```diff
@@ -195,21 +195,21 @@ typedef int32_t WPMediaRequestID;
  *
  *  @return an object implementing WPMediaGroup protocol.
  */
-- (id<WPMediaGroup>)groupAtIndex:(NSInteger)index;
+- (nonnull id<WPMediaGroup>)groupAtIndex:(NSInteger)index;
```
frosty: Is it possible that this could return `nil` if the index is out of bounds?

Pod/Classes/WPMediaCollectionDataSource.h
```diff
@@ -225,7 +225,7 @@ typedef int32_t WPMediaRequestID;
  *
  *  @return an object implementing the WPMediaAsset protocol
  */
-- (id<WPMediaAsset>)mediaAtIndex:(NSInteger)index;
+- (nonnull id<WPMediaAsset>)mediaAtIndex:(NSInteger)index;
```
frosty: Same question as for `groupAtIndex:`?

SergioEstevao: I was thinking if you use an index out of bounds to throw an exception, like the system array.

PR #235, opened 2017-09-06
Fixes content offset issue when the inline picker is set to vertical

bummytime:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/7754

napkin 47 09-06-17 2 50 30 pm

This PR fixes an issue where the collection view's content offset is modified because  `viewDidLayoutSubviews` is called multiple times before the photo grid is displayed. Now, we only create the flow layout once and then modify it accordingly in `configureCollectionView`.

H/T: [This guy](http://samwize.com/2016/04/04/the-bug-to-do-with-uicollectionview-content-offset/#comment-2896704421). 

### To test

In the media picker demo:
1. Set the input picker to scroll vertically (in the options)
2. Tap on the media picker quick select field — verify the content is not cut off.
3. Long press on any picture to preview and then tap cancel on the preview VC.
4. When the input picker appears again, verify the content is not cut off.

Test in WPiOS using [this branch](https://github.com/wordpress-mobile/WordPress-iOS/tree/issue/7754-media-picker-scroll-fix):
1. Open the Aztec editor to start a new post.
2. Select the + to open the inline media picker. Note that it loads as expected (at the very top of the list of device photos, starting with most recent).
3. Select a different media source (device photo library, camera, or WordPress media library).
4. Cancel to go back to the inline picker.
5. Verify the top row is not cut off (as described in the original bug).

Repeat a few times in WPiOS, however, try scrolling to random points in the inline picker before choosing a different media source. After canceling, verify the input picker returns to the exact same scroll point.

@SergioEstevao would you mind taking a quick look?

### DIFF
--- a/Pod/Classes/WPInputMediaPickerViewController.m
+++ b/Pod/Classes/WPInputMediaPickerViewController.m
@@ -59,6 +59,7 @@ static CGFloat const IPadPro12LandscapeWidth = 1366.0f;
     
     self.mediaPicker.view.frame = self.view.bounds;
     self.mediaPicker.view.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+    self.mediaPicker.collectionView.collectionViewLayout = [[UICollectionViewFlowLayout alloc] init];
     [self.view addSubview:self.mediaPicker.view];
     [self.mediaPicker didMoveToParentViewController:self];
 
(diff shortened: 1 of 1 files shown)


SergioEstevao:
@bummytime  I went ahead and merge this in to get it on the 0.21 release.

bummytime:
@SergioEstevao Thank you!

PR #238, opened 2017-09-08
Refactor layout code to be all inside WPMediaPickerViewController.

SergioEstevao:
This PR refactor the code to set layout and scrolling horizontally or vertically to all be part of `WPMediaPickerViewController` instead of being split between `WPInputMediaViewController` and `WPMediaPickerViewController`.

It also moves the `scrollDirection` option to inside `WPMediaPickerOptions` to make it more consistent.

To test:
 - Launch the demo app
 - Check the input picker and the full screen picker on vertical and horizontal scroll.

### DIFF
--- a/Pod/Classes/WPInputMediaPickerViewController.m
+++ b/Pod/Classes/WPInputMediaPickerViewController.m
@@ -1,13 +1,6 @@
 #import "WPInputMediaPickerViewController.h"
 #import "WPPHAssetDataSource.h"
 
-static CGFloat const IPhoneSELandscapeWidth = 568.0f;
-static CGFloat const IPhone7PortraitWidth = 375.0f;
-static CGFloat const IPhone7LandscapeWidth = 667.0f;
-static CGFloat const IPadPortraitWidth = 768.0f;
-static CGFloat const IPadLandscapeWidth = 1024.0f;
-static CGFloat const IPadPro12LandscapeWidth = 1366.0f;
-
 @interface WPInputMediaPickerViewController()
 
 @property (nonatomic, strong) WPMediaPickerViewController *mediaPicker;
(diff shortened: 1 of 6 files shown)


Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -309,6 +353,7 @@ - (void)refreshDataAnimated:(BOOL)animated
             strongSelf.collectionView.allowsSelection = YES;
             strongSelf.collectionView.allowsMultipleSelection = strongSelf.options.allowMultipleSelection;
             strongSelf.collectionView.scrollEnabled = YES;
+            strongSelf.collectionView.bounces = YES;
```
frosty: `YES` is the default, so do we need this here?

Pod/Classes/WPInputMediaPickerViewController.m
```diff
@@ -73,85 +65,6 @@ - (void)setupMediaPickerViewController {
...
-    if (self.scrollVertically) {
-        CGFloat frameWidth = self.view.frame.size.width;
-        NSUInteger numberOfPhotosForLine = [self numberOfPhotosPerRow:frameWidth];
-
-        photoSize = [self.mediaPicker cellSizeForPhotosPerLineCount:numberOfPhotosForLine
-                                                       photoSpacing:photoSpacing
-                                                         frameWidth:frameWidth];
-
-        // Check the actual width of the content based on the computed cell size
-        // How many photos are we actually fitting per line?
```
frosty: What happened to this code? I don't see this section in the updated `WPMediaPickerCollectionView`?

SergioEstevao: I realised that we didn't need to do this any longer, because the way we calculate the number of rows.

Pod/Classes/WPMediaPickerViewController.m
SergioEstevao: This can go away, I was try to find why our collection view doesn't bounce...

PR #248, opened 2017-10-03
Update project to be iOS10 and above compatible.

SergioEstevao:
This PR makes the pod to required iOS10 as a minimum and makes all the necessary updates to the codebase. This also updates the project to the latest Xcode 9 recommended settings.

To test:
 - Make sure the project compiles without warnings.
 - Check the following scenarios where changes where made:
  - appearance of the cells on the picker
  - block the permissions of the app to access the Photos album and check if the dialog that gives access to Setting work correctly
  - Check if the front camera is selected when the option is activated.


bummytime:
Looks great @SergioEstevao.

✅ Make sure the project compiles without warnings.
✅ appearance of the cells on the picker
✅ block the permissions of the app to access the Photos album and check if the dialog that gives access to Setting work correctly
✅ Check if the front camera is selected when the option is activated.

Nice work!

PR #252, opened 2017-10-24
Issue/230 hide empty albums

### LINKED ISSUE #230: Hide empty albums

SergioEstevao:
Fixes #230 

This doesn't filter completely out by media type, but at least tries to remove albums that are completely empty.

To test:
 - Launch the demo app
 - Open the full screen picker
 - Check that empty albums are not showed
 - Check that is not slow.


SergioEstevao:
@frosty fixed the crash can you give it another look.

PR #260, opened 2017-11-08
Feature/search bar

### LINKED ISSUE #257: Add a search option to the Picker
It would be nice to allow users search for the media, something like this:

screen shot 2017-10-31 at 14 29 50

Related to this request on the media app: https://github.com/wordpress-mobile/WordPress-iOS/issues/7680

This is basically to allows us to have Search possibilities anywhere we use the picker.

etoledom:
Fixes #257 

- The search bar will be displayed via `WPMediaPickerOptions` property.
- The data source must implement a new method for an opportunity to filter elements.

I tried to implement searching and filtering with the default `WPPHAssetDataSource` provided by this component and to show it in the example project without much luck.
The predicate options available for `PHAsset` objects are limited to a few properties as explained [here](https://developer.apple.com/documentation/photos/phfetchoptions), and none of them are usable for searching.

### DIFF
--- a/Pod/Classes/WPMediaCollectionDataSource.h
+++ b/Pod/Classes/WPMediaCollectionDataSource.h
@@ -320,7 +320,7 @@ typedef int32_t WPMediaRequestID;
 @optional
 
 /**
- *  Tells the Data Source that the search string has changed
+ *  Tells the Data Source that the search string has been changed
  *
  *  @param searchText the new search text
  */
--- a/Pod/Classes/WPMediaCollectionDataSource.h
+++ b/Pod/Classes/WPMediaCollectionDataSource.h
@@ -317,5 +317,14 @@ typedef int32_t WPMediaRequestID;
  */
 - (BOOL)ascendingOrdering;
 
+@optional
+
+/**
+ *  Tells the Data Source that the search string has changed
+ *
+ *  @param searchText the new search text
+ */
+- (void)searchTextDidChange:(nullable NSString *)searchText;
+
 @end
 
--- a/Pod/Classes/WPMediaCollectionDataSource.h
+++ b/Pod/Classes/WPMediaCollectionDataSource.h
@@ -324,7 +324,7 @@ typedef int32_t WPMediaRequestID;
  *
  *  @param searchText the new search text
  */
-- (void)searchTextDidChange:(nullable NSString *)searchText;
+- (void)searchFor:(nullable NSString *)searchText;
 
 @end
 
--- a/Pod/Classes/WPMediaPickerViewController.m
+++ b/Pod/Classes/WPMediaPickerViewController.m
@@ -305,10 +305,11 @@ static CGFloat SelectAnimationTime = 0.2;
 
 - (void)setupSearchBar
 {
-    if ([self.parentViewController isKindOfClass:[WPInputMediaPickerViewController class]]) {
-        return; // To not allow search bar on WPInputMediaPicker
-    }
-    if (self.options.showSearchBar && self.searchBar == nil) {
+    BOOL showldShowSearchBar = self.options.showSearchBar &&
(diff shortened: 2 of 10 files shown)


Pod/Classes/WPMediaCollectionDataSource.h
```diff
@@ -317,5 +317,14 @@ typedef int32_t WPMediaRequestID;
...
 - (BOOL)ascendingOrdering;
 
+@optional
+
+/**
+ *  Tells the Data Source that the search string has been changed
+ *
+ *  @param searchText the new search text
+ */
+- (void)searchTextDidChange:(nullable NSString *)searchText;
```
SergioEstevao: I'm not a fan of this method name in the data source I'm more inclined to something like `filterByTerm` or `searchFor` this looks too associated to the delegate callback of the searchBar.

etoledom:
@SergioEstevao Thanks for your review! I’ll make all the requested changes and additions today.

As for the PHAsset predicate, I tried using `SELF` / `description` to get some usable string without any positive result. 

The `creationDate` might be an option too. Another way could be to use PHCollection  for filtering? In that case we could use the `localizedTitle` property directly, but this would make more sense in the Albums list. In any case I haven’t studied the data source implementation thoroughly yet, so I can’t be sure if this idea would work. 🤔 I´ll study this in more detail. 

Do you have any suggestions?

SergioEstevao:
@etoledom Good work on the changes.

Regarding the searchbar for PHAsset, me and @frosty where discussing the use of the creationDate, we could check if the use wrote something that was a date (NSDateDetector) and then filter assets around that. But no need to. implement this on this PR. We can do a separate one for just that.

But I do think for cases where the DataSource is not searching anything, and because we can mix multiple DataSource (like we do in the main app) it will be good to add another method to the protocol called `isSearchable` that will indicate if the dataSource is searchable and show/hide the searchbar accordingly.

etoledom:
@SergioEstevao Thanks! :]

With regards to the option you’ve proposed, may I suggest an alternative solution? What do you think about checking if the data source overrides the `searchFor` method and using it as the “isSearchable” information (since it’s optional)?

This way to show a searchBar:
- The `showSearchBar` media picker option must be set to `true`.
- The given data source must override the `searchFor:` method, making it “searchable” (This would have to be implemented).

And to not show a search bar we simply set the the option to `false`, independently of the data source.

The benefit of this approach is that we can keep the API simple, getting the same information without an extra methods to override.

How does that sound?

SergioEstevao:
@etoledom I like your solution, go for it.

etoledom:
I have a small question: since now it´s required to override `searchFor:` to show the searchBar, and the `WPPHAssetDataSource` is not overriding it yet, the searchBar won’t be visible in the demo app even if the option is set to show it. Is that fine for now or do you prefer a dummy override so the searchBar can be visible?

etoledom:
Since a `searchFor:` override  is necessary to show the searchBar and the `WPPHAssetDataSource` is not filtering yet, I made a dummy override so the search bar can be shown and tested in the demo app. If you prefer it disabled I can do it quickly to have this PR merged before the end of the week. 😁

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -305,10 +305,11 @@ - (void)addCollectionViewToView
 
 - (void)setupSearchBar
 {
-    if ([self.parentViewController isKindOfClass:[WPInputMediaPickerViewController class]]) {
-        return; // To not allow search bar on WPInputMediaPicker
-    }
-    if (self.options.showSearchBar && self.searchBar == nil) {
+    BOOL showldShowSearchBar = self.options.showSearchBar &&
```
SergioEstevao: Small typo `should`instead of `showld`

Issue #259, opened 2017-11-06
Allow cells to include an overlay provided by the presenting application


frosty:
We'd like to add an overlay view to media picker cells, provided by the application using the library. In WordPress for iOS, we'll use this to add upload progress indicators and retry buttons.

![async-cross-i7](https://user-images.githubusercontent.com/4780/32451347-7fb6c602-c30e-11e7-8e62-e37a1efccbc1.png)
![async-cross-i7-2](https://user-images.githubusercontent.com/4780/32451288-646cb032-c30e-11e7-8f9a-a1988d2bb791.png)

PR #261, opened 2017-11-09
Add configurable overlay views to cells

frosty:
Implements #259.

This PR adds a configurable 'overlay' view to cells in the picker. It does so via two new delegate methods on the picker:

```
- (BOOL)mediaPickerController:(WPMediaPickerViewController *)picker shouldShowOverlayViewForCellForAsset:(id<WPMediaAsset>)asset;

- (void)mediaPickerController:(WPMediaPickerViewController *)picker willShowOverlayView:(UIView *)overlayView forCellForAsset:(id<WPMediaAsset>)asset;
```

If the first method isn't implemented, or returns `NO`, no overlays will be shown. The second method gives the delegate an opportunity to customise the overlay view for the specific cell. The overlay view is also exposed as a property on `WPMediaCollectionViewCell`, to allow customisation at other times (for example when an upload completes).

By default, the overlay view will be a plain `UIView`. This can be customised through a new method on `WPMediaPickerController`:

`- (void)registerClassForReusableCellOverlayViews:(Class)overlayClass`
 
If you pass a class to this method when configuring the picker, instances of that class will be instantiated when a new overlay view is needed.

Overlay views are recycled, so that we only create as many as we need for cells displayed onscreen, and aren't constantly recreating them.

I've also added a new option to the demo app where you can toggle on some sample overlays for the cells there.

simulator screen shot - iphone 8 - 2017-11-09 at 11 20 25

To test:

* Build and run the demo app
* Turn on the new "Show Sample Cell Overlays" option
* Present the picker, and check that you can see the filename shown at the top of the cells
* Ensure you have multiple screens of media items in the picker, and try scrolling through them quickly and check that the overlays show correctly.
* Also in the above situation, try using Xcode's memory graph debugger to check that we only ever create a limited number of cells (should be around 28 on an iPhone 8 sized screen).

Needs review: @SergioEstevao

### DIFF
--- a/Pod/Classes/WPMediaPickerViewController.m
+++ b/Pod/Classes/WPMediaPickerViewController.m
@@ -36,6 +36,10 @@ static CGFloat const IPadPro12LandscapeWidth = 1366.0f;
 @property (nonatomic, assign) BOOL refreshGroupFirstTime;
 @property (nonatomic, strong) UILongPressGestureRecognizer *longPressGestureRecognizer;
 @property (nonatomic, strong) NSIndexPath *assetIndexInPreview;
+
+@property (nonatomic, strong, nullable) Class overlayViewClass;
+@property (nonatomic, strong) NSArray<UIView *> * reusableOverlayViews;
+
 /**
  The size of the camera preview cell
  */
@@ -151,6 +155,15 @@ static CGFloat SelectAnimationTime = 0.2;
             [self.collectionView reloadData];
         }
     }
+}
+
+- (void)registerClassForReusableCellOverlayViews:(Class)overlayClass
+{
+    NSParameterAssert([overlayClass isSubclassOfClass:[UIView class]]);
+
+    self.overlayViewClass = overlayClass;
+
+    self.reusableOverlayViews = @[];
 }
 
 - (UICollectionViewFlowLayout *)layout
@@ -494,8 +507,14 @@ static CGFloat SelectAnimationTime = 0.2;
 
 - (UIView *)dequeueReusableOverlayView
 {
-    UIView *view = [UIView new];
-    view.backgroundColor = [UIColor colorWithWhite:1.0 alpha:0.3];
+    UIView *view = nil;
+
+    if (self.overlayViewClass) {
+        view = [self.overlayViewClass new];
+    } else {
+        view = [UIView new];
+    }
+
     return view;
 }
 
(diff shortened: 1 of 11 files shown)


Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -36,6 +36,10 @@ @interface WPMediaPickerViewController ()
 @property (nonatomic, assign) BOOL refreshGroupFirstTime;
 @property (nonatomic, strong) UILongPressGestureRecognizer *longPressGestureRecognizer;
 @property (nonatomic, strong) NSIndexPath *assetIndexInPreview;
+
+@property (nonatomic, strong, nullable) Class overlayViewClass;
+@property (nonatomic, strong) NSMutableArray<UIView *> * reusableOverlayViews;
```
SergioEstevao: @frosty do we really need to have our own reusableOverlayViews storage? couldn't create the overlayView we create the dequeue the WPMediaCollectionViewCell? 

So something like this:

```
WPMediaCollectionViewCell *cell = [collectionView dequeueReusableCellWithReuseIdentifier:NSStringFromClass([WPMediaCollectionViewCell class]) forIndexPath:indexPath];

if (cell.overlayView == nil || ![cell.overlayView isKindOfClass: overlayViewClass]){
  cell.overlayView = [self.overlayViewClass new];
}
```

This way we will just reuse the cache for cells on the CollectionView, no need to implement ours.

What do you think?

frosty: So I'm doing our own storage because not every cell might have an overlay view. If the delegate returns `NO` to `shouldDisplayOverlayView`, we won't create one at all. This means if we only have one item uploading, for example, we'd only create one overlay view, instead of having ~28 in cells that we're not using. Does that make sense?

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -492,6 +510,22 @@ - (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSe
...
+
+    if (view) {
+        [self.reusableOverlayViews removeLastObject];
+        return view;
+    }
+
+    if (self.overlayViewClass) {
+        return [self.overlayViewClass new];
+    } else {
+        return [UIView new];
```
SergioEstevao: We probably should throw an error if no class is configured and we need to use one.

SergioEstevao:
Excellent work @frosty !

frosty:
@SergioEstevao conflicts fixed – do you want to double check? I'm not seeing the search bar anywhere...

SergioEstevao:
@frosty conflicts again... do you want to give it another go?

frosty:
@SergioEstevao Fixed!

SergioEstevao:
@frosty can I merge or you want do it?

PR #262, opened 2017-11-10
Use system strings for relative dates.

SergioEstevao:
This PR refactor the code for WPDateTimeHelpers to use system functionality for relative dates instead of manual coding the strings.

### DIFF
--- a/Pod/Classes/WPDateTimeHelpers.m
+++ b/Pod/Classes/WPDateTimeHelpers.m
@@ -1,5 +1,8 @@
 #import "WPDateTimeHelpers.h"
 
+// To use only on unit tests to check results in specific languages
+static NSString *forcedLocaleIdentifier = nil;
+
 @implementation WPDateTimeHelpers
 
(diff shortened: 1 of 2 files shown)


Pod/Classes/WPDateTimeHelpers.m
```diff
@@ -1,5 +1,8 @@
 #import "WPDateTimeHelpers.h"
 
+// To use only on unit tests to check results in specific languages
+static NSString *forcedLocaleIdentifier = nil;
```
frosty: I might be misunderstanding, but why do we need this `static` property? Can't we just have the `setForcedLocaleIdentifier` method set the locale directly on the shared formatter? Are we using this for anything?

SergioEstevao: You are right! I started doing this in another way, then continued to used the variable anyway when switching to this approach. Will fix this.

SergioEstevao: @frosty ready for another look.

PR #265, opened 2017-11-11
Fix search bar inset issues

### LINKED ISSUE #264: Search bar inset issues
I've noticed a couple of issues with the new search bar in the media picker:

The bottom scroll / content insets of the collection view don't seem to be adjusted when the keyboard is shown:

simulator screen shot - iphone 8 - 2017-11-10 at 14 39 34

Notice the content scrolls down behind the keyboard, and the scroll indicator is hidden behind there too.

Also, the search bar is inset on both sides on the iPhone X in landscape, but I'd expect its background to stretch to the edges of the screen. Compare the current search bar in the media library in WPiOS:

screen shot 2017-11-10 at 20 20 29

with the search bar in the media picker:

screen shot 2017-11-10 at 20 17 21

cc @etoledom

etoledom:
This PR fixes some issues with the picker content inset (and scroll indicator inset) when the keyboard is shown or hidden.
It also makes the picker view and search bar use the full width in the landscape mode on iPhone X.
Finally, a small fix was added to the initializer of the `WPMediaPickerViewController` to avoid a crash.

Fixes #264 

To test:

- Build and run the demo app.
- Present the picker and show the keyboard by tapping on the search bar.
- Scroll until the last element.
- The last element and the scroll indicator should be completely visible.
- Cancel the search to dismiss the keyboard.
- The last element should still be completely visible.

img_1927
All this should also work in the presence of tab bars (opaque and translucent).

Test iPhone X spaces:
- On the iPhone X in landscape mode, the search bar and the scroll view background should use the full width of the screen.

screen shot 2017-11-11 at 6 36 22 pm

### DIFF
--- a/Pod/Classes/WPNavigationMediaPickerViewController.m
+++ b/Pod/Classes/WPNavigationMediaPickerViewController.m
@@ -288,13 +288,6 @@ static NSString *const ArrowDown = @"\u25be";
     }
 }
 
-#pragma mark - UIPopoverPresentationControllerDelegate
-
--(void)prepareForPopoverPresentation:(UIPopoverPresentationController *)popoverPresentationController
-{
-    self.mediaPicker.presentedAsPopover = YES;
-}
-
 #pragma mark - Public Methods
 
 - (void)showAfterViewController:(UIViewController *)viewController
--- a/Pod/Classes/WPNavigationMediaPickerViewController.m
+++ b/Pod/Classes/WPNavigationMediaPickerViewController.m
@@ -53,8 +53,7 @@ static NSString *const ArrowDown = @"\u25be";
 - (void)viewDidLoad
 {
     [super viewDidLoad];
-
-    self.popoverPresentationController.delegate = self;
+    
     self.view.backgroundColor = [UIColor whiteColor];
 
     [self setupNavigationController];
--- a/Pod/Classes/WPNavigationMediaPickerViewController.m
+++ b/Pod/Classes/WPNavigationMediaPickerViewController.m
@@ -54,6 +54,7 @@ static NSString *const ArrowDown = @"\u25be";
 {
     [super viewDidLoad];
 
+    self.popoverPresentationController.delegate = self;
     self.view.backgroundColor = [UIColor whiteColor];
 
     [self setupNavigationController];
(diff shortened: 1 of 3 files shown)


SergioEstevao:
@etoledom it looks there is still a problem with the iPad using a popover. It looks that when in popover mode it's taking in account the keyboard size. Here is a screenshot:

screen shot 2017-11-13 at 10 48 07

SergioEstevao:
@etoledom there is still an edge case that is failing, so If you start with the iPad like the screenshot above is displaying well, but then if you start multitask and switch the view mode to a compact mode you get this:

screen shot 2017-11-14 at 17 53 26

So the insets are now wrong because they are ignoring the keyboard.

I think you will need to have a more dynamic approach detecting the state of the popover mode.
Check this article: http://petersteinberger.com/blog/2015/uipresentationcontroller-popover-detection/

etoledom:
@SergioEstevao Thanks for sharing the article!

I knew about the method of using `_UIPopoverView`, but since it is checking for private API I tried to avoid it. But after reading the article and testing some other options I figured that what definitely works and is quite simple is the check for `_UIPopoverView` in the view hierarchy. 

From what I read it seems to pass the Apple app review process but it’s uncertain if it will break with the future iOS releases.

Should we go with this one?

SergioEstevao:
@etoledom Did you try the method that check for the arrow direction, it's referred on the end of the article? If that work ok it's an valid option without the need of using private APIs.

etoledom:
@SergioEstevao I gave it a try in the afternoon and it just returned me a bunch of zeroes for both popover and modal presentations. Now it seems to be working just fine, so I’m not sure what happened there. 🤔

I’ll do some more testing and if everything works fine I will implement it and push it in a moment.

etoledom:
@SergioEstevao I just sent the fix we talked about. I found why it failed when I tested it the first time.
In iOS 11 the popoverPresentationController seems to propagate through the controller hierarchy, so the arrow direction was easily found in the picker controller itself. But in iOS 10 this doesn’t happen and I was getting 0s (nil). 
To fix that, a recursive search is now being performed to find the nearest popoverPresentationController in the hierarchy, if any.

I also found a couple of extra issues related to the keyboard when the picker is set in horizontal mode:
- The height of the image capture cell wasn’t taking into account the space taken by the keyboard.
- The collection view layout wasn’t refreshing when the keyboard appeared / dissappeared until a scroll action was performed.

All these issues should be fixed now. 🙃

Pod/Classes/WPNavigationMediaPickerViewController.m
```diff
@@ -54,6 +54,7 @@ - (void)viewDidLoad
 {
     [super viewDidLoad];
 
+    self.popoverPresentationController.delegate = self;
```
SergioEstevao: Do we still need this here?

etoledom: It's not needed anymore. My mistake 😅

PR #268, opened 2017-11-17
Added extra API needed to know about specific events related to the search bar

### LINKED ISSUE #267: Extra API needed to know about specific events related to the search bar.
I would be good to have a couple of extra methods in the search API:

- A method to let the data source know when the user cancels the search, so it has an opportunity to clean up any search-related state.
- A method to let the delegate know that the amount of elements displayed changed by a search query. It could be useful to let the controller react to the new element count.

etoledom:
Fixes #267 

After testing the implementation of the new search bar in the main WPiOS app I realized it was necessary to add five extra methods to match the actual behavior of this search bar to that of the one in the media library. 

This PR adds these extra methods:

- `searchCancelled` method in the data source to clean search states. 
Used to clean search queries in MediaLibraryPickerDataSource and re-fetch data.

- `didUpdateSearchWithAssetCount:` method in the media picker delegate.
Used to refresh the empty view state.

- `emptyViewForMediaPickerController:` method in the media picker delegate.
Used to display the `WPNoResultsView`. The media picker library also manages the empty view center animation when the keyboard appears / disappears.

- `showSearchBar` / `hideSearchBar` methods in `WPMediaPickerViewController`.
Used to show / hide the search bar when necessary.

Here is a gif that shows this implementation working in WPiOS
It also shows the search bar implemented in the editor picker with the default empty view:

nov-16-2017 23-28-14

Since the current implementation of the `WPPHAssetDataSource` doesn’t have a proper support for search yet, it’s not possible to test this new implementation graphically in the demo app as it is now.

Needs review: @SergioEstevao

### DIFF
--- a/Pod/Classes/WPMediaPickerViewController.m
+++ b/Pod/Classes/WPMediaPickerViewController.m
@@ -17,8 +17,6 @@ static CGFloat const IPadPortraitWidth = 768.0f;
 static CGFloat const IPadLandscapeWidth = 1024.0f;
 static CGFloat const IPadPro12LandscapeWidth = 1366.0f;
 
-static NSString * const kDefaultEmptyText = @"Nothing to show";
-
 @interface WPMediaPickerViewController ()
 <
  UICollectionViewDataSource,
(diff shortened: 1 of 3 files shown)


Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -17,6 +17,8 @@
 static CGFloat const IPadLandscapeWidth = 1024.0f;
 static CGFloat const IPadPro12LandscapeWidth = 1366.0f;
 
+static NSString * const kDefaultEmptyText = @"Nothing to show";
```
SergioEstevao: This string needs to be localised, just use NSLocalizedString() with a an adequate comment, and our translation scripts will pick this up.

etoledom:
Ready! I removed the constant too, following the pattern found in the main WPiOS app.

PR #269, opened 2017-11-17
WPMediaPickerViewController: Reconfigure (don't reload) cells on update

frosty:
This PR updates the change handling in `WPMediaPickerViewController` to match Apple's [sample code](https://developer.apple.com/documentation/coredata/nsfetchedresultscontrollerdelegate) for fetched results controller changes. Instead of reloading cells whenever their associated objects change, we just fetch the existing cell and update its contents. This prevents the collection view recreating cells entirely, which is useful when updating upload progress in WPiOS.

To test:

* Build and run, and ensure that cells update correctly when their contents change – it may be easiest to use this branch from WPiOS, and try uploading some items.


frosty:
Thanks!

PR #272, opened 2017-11-21
Adds support for Smart Invert accessibility option

frosty:
iOS 11 introduces a new accessibility feature called 'Smart Invert Colors'. This is like the old 'invert colors' feature, but can exclude specific content from being inverted – instead it's displayed slightly dimmed or desaturated.

This PR sets image views within MediaPicker-iOS that display photographic content so they're not inverted when Smart Invert Colors is enabled. Unfortunately the feature doesn't screenshot correctly (the entire UI is inverted, not just the images), but this gives you a rough idea of what the PR fixes:

smart-invert

To test:

* On an iOS 11 device, enable Smart Invert Colors – go into Settings > General > Accessibility > Display Accommodations > Invert Colors > turn on Smart Invert.
* View the groups list, the collection view inside a group, and the asset preview for a single asset. Ensure that their colours aren't inverted.

### DIFF
--- a/Pod/Classes/WPAssetViewController.m
+++ b/Pod/Classes/WPAssetViewController.m
@@ -103,6 +103,9 @@
     _imageView.backgroundColor = [UIColor blackColor];
     _imageView.userInteractionEnabled = YES;
     [_imageView addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(handleTapOnAsset:)]];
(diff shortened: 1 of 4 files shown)


frosty:
Just noticed I missed the video view, so I'll push an update for that soon.

frosty:
@SergioEstevao I added the same for the video player if you want to check that.

SergioEstevao:
Looking good with videos!

PR #273, opened 2017-11-21
Video player: Toolbar / home indicator fixes for iPhone X

frosty:
This PR makes a few tweaks to the toolbar in the video player to better support iPhone X.

* The home indicator no longer overlaps the toolbar
* The home indicator is now hidden automatically when the toolbar is hidden

picker-video

To test:

* In the demo app on iPhone X, open the picker and find a video. Open the video detail by 3D touching / long pressing on the video thumbnail.
* Tap to show the toolbar.
* Ensure that the toolbar controls sit above the home indicator in both orientations.
* Tap the screen again to hide the toolbar. Ensure the the home indicator is hidden after a brief delay.
* Tap again to show the toolbar, and ensure that the home indicator re-appears.
* Ensure that it still looks right on a non-iPhone X device, and that it doesn't crash on iOS 10.

### DIFF
--- a/Pod/Classes/WPAssetViewController.m
+++ b/Pod/Classes/WPAssetViewController.m
@@ -201,16 +201,29 @@
     }
 }
 
-- (BOOL)prefersStatusBarHidden {
+- (BOOL)prefersStatusBarHidden
+{
+    return self.videoView.controlToolbarHidden;
+}
+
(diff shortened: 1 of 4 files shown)


frosty:
Thanks Sergio!

PR #276, opened 2017-12-26
Support limiting the number of selected assets

shahla00:
Fixes #

To test:


SergioEstevao:
@shahla00 Thanks for the PR, the code is looking great, but I would like two changes to be done before this is merged:

- Keep the allowMultipleSelection property separated from numberOfAssetsLimit
- Create a constant that allows infinite selection

Thanks again

PR #278, opened 2018-02-09
Move selection indicator and frame in front of overlay view

frosty:
As part of allowing batch deletion of failed media in WPiOS, we need to allow those failed cells to be selected. At the moment, the overlay view added to failed cells appears on top of the selection indicator and frame, making them less visible. This PR switches the order of things around.

**Before:**

screen shot 2018-02-09 at 16 37 29

**After:**

screen shot 2018-02-09 at 16 31 51

**To test:**

* Build and run the demo app, and enable the sample overlay cells in options. Then tap the + in the top right corner to bring up the picker. Tap multiple cells to select them, and ensure the selection indicator and frame are visible.


frosty:
Thanks!

PR #281, opened 2018-04-02
Implement a carousel view to preview multiple selection of assets

etoledom:
This PR adds implements a carousel assets view to preview assets selected under multiple selection.

The [original design](https://user-images.githubusercontent.com/2722505/37941996-994ad1ea-31a3-11e8-9660-19badea57e28.jpg) was made for Stock Photos, and the original issue can be found [here](https://github.com/wordpress-mobile/WordPress-iOS/issues/8954).

Work In progress:

- [x] Implement base pagination
- [x] Reconnect delegates
- [x] Add multiple / single preview option to sample project
- [x] Documentation

To test:
1. Open the picker.
2. Select multiple assets.
3. Open preview via force touch (or long press in the Input view).
4. Scroll and see all your assets.

### DIFF
--- a/Pod/Classes/WPCarouselAssetsViewController.m
+++ b/Pod/Classes/WPCarouselAssetsViewController.m
@@ -28,7 +28,7 @@
     [self updateTitle];
 }
 
-- (void)setIndex:(NSInteger)index animated:(BOOL)animated
+- (void)setPreviewingAssetAtIndex:(NSInteger)index animated:(BOOL)animated
 {
     self.index = index;
     if (self.isViewLoaded) {
@@ -100,8 +100,9 @@
 
 - (void)updateTitle
 {
-    NSString *separator = NSLocalizedString(@"of", @"Word separating the current index of the total amount. I.e.: 7 of 9");
-    self.title = [NSString stringWithFormat:@"%ld %@ %ld", self.index + 1, separator, self.assets.count];
+    NSString *separator = NSLocalizedString(@"of", @"Word separating the current index from the total amount. I.e.: 7 of 9");
+    long showingCount = self.index + 1;
+    self.title = [NSString stringWithFormat:@"%ld %@ %ld", showingCount, separator, (long)self.assets.count];
 }
 
 - (NSInteger)indexForViewController:(UIViewController *)viewController
--- a/Pod/Classes/WPCarouselAssetsViewController.m
+++ b/Pod/Classes/WPCarouselAssetsViewController.m
@@ -37,7 +37,7 @@
     }
 }
 
-#pragma mark: - UIPageViewControllerDelegate
+#pragma mark - UIPageViewControllerDelegate
 
 - (UIViewController *)pageViewController:(UIPageViewController *)pageViewController viewControllerBeforeViewController:(UIViewController *)viewController
 {
@@ -57,7 +57,7 @@
     return [self viewControllerAtIndex:index + 1];
 }
 
-#pragma mark: - UIPageViewControllerDelegate
+#pragma mark - UIPageViewControllerDelegate
 
 - (void)pageViewController:(UIPageViewController *)pageViewController didFinishAnimating:(BOOL)finished previousViewControllers:(NSArray<UIViewController *> *)previousViewControllers transitionCompleted:(BOOL)completed
 {
@@ -73,7 +73,7 @@
     self.nextIndex = [self indexForViewController:nextViewController];
 }
 
-#pragma mark: - WPAssetViewControllerDelegate
+#pragma mark - WPAssetViewControllerDelegate
 
 - (void)assetViewController:(nonnull WPAssetViewController *)assetPreviewVC failedWithError:(nonnull NSError *)error {
     if (self.assetViewDelegate) {
@@ -87,7 +87,7 @@
     }
 }
 
-#pragma mark: - Helpers
+#pragma mark - Helpers
 
 - (void)initialSetup
 {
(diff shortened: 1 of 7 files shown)


etoledom:
@SergioEstevao or @frosty , would you mind taking a fast look at this work in progress?

Firstly I thought on creating a custom preview in the main app. But since I would have to modify the Media Picker delegate anyway, I though on extending the default preview in the library itself.

In this way we get free multiple preview everywhere in the main app that uses multiple selection without changing anything. I think that is a win.

Implementation wise, I'm not sure if it's better to leave just the new delegate method:
`- (nullable UIViewController *)mediaPickerController:(nonnull WPMediaPickerViewController *)picker previewViewControllerForAssets:(nonnull NSArray<id<WPMediaAsset>> *)assets selectedIndex:(NSInteger)selected;`
Or to keep both.

I'm also not sure if it worth it to modify the example app to add multiple / single preview options.

Also any idea for improvement or desired functionality are welcome :)

Thanks in advance!

frosty:
@etoledom This is looking good! I think it makes sense to modify the Media Picker itself to allow this :)

* I think for now I would keep both delegate methods, but mark the 'old' way as `deprecated`.
* Although now I think about it, could we not keep `previewViewControllerForAsset:` and we just get the media picker view controller's `selectedAssets` in the implementation of the delegate method?
* I would just keep the ability to use a custom preview VC in the example app, but don't worry specifically about multiple / single preview options.

etoledom:
@frosty Thanks for the review!

I pushed some more commits following your comments. The second point you mentioned I didn't really get it. Would you mind to explain it a bit more?

I mixed the single asset preview with the carousel preview in this way:
- If I selected multiple assets (more than one) and I try to preview one of them. It will show the carousel preview focused on the asset I wanted to preview.
- If I preview an asset that is not selected, it will preview that single asset, independently of how many other assets are selected.
- If I have selected one asset and I want to preview that one, it will open the single preview.

[There is a demonstration video here.](https://www.dropbox.com/s/kh7v2054tofoj2u/media_01.mp4?dl=0)

There is also a detail with Audio files (that won't affect Stock Photos). It's managed in this way:
- Trying to preview audio files won't work for any circumstance.
- When there are many assets selected, including audio files, why I preview one of those non-audio files, will show the carousel preview of all non-audio files, striping out all audio files.

[There's a video of that here](https://www.dropbox.com/s/d0rp8lz4g6lmun0/media_02.mp4?dl=0)

@iamthomasbishop There is a nice small detail in the old single asset preview:
At the top right corner, there is an `Add / Remove` button that adds or remove that asset to the list of selected assets (the first video demonstrate it). Do you think it is worth it to add it to the carousel preview too?

frosty:
Hey @etoledom! The way you've missed single and multiple asset previews, and the handling of audio files all makes perfect sense to me 👍 

Regarding my comment:

> I would just keep the ability to use a custom preview VC in the example app, but don't worry specifically about multiple / single preview options.

I might have misunderstood your earlier comment: "I'm also not sure if it worth it to modify the example app to add multiple / single preview options". I thought you were wondering whether to add options to the example app to allow the user to show either single or multiple previews? I was just saying that I don't think that's necessary, but I would leave in place the current option that uses a custom preview view controller:

screen shot 2018-04-04 at 09 46 14

frosty:
@etoledom When you're ready for a 'proper' review, just let me know / add me to the PR :)

etoledom:
Thanks @frosty !
The `Use custom preview controller` is working with the new delegate 👍

Let's wait for @iamthomasbishop comments to know if there's anything else to do :)

iamthomasbishop:
@etoledom The carousel flow in general is looking fantastic. Well done!

Regarding audio files - I think we should still show them in the carousel, with the same display style as on the grid (same dark-blue background, filename, and icon). We don't necessarily need to allow the user to play the audio file, but it would be nice if it were included in the carousel while they've selected it (it'd be odd to me if I've selected 4 images and 2 audio files, if I only saw 4 images).

> There is a nice small detail in the old single asset preview: At the top right corner, there is an Add / Remove button that adds or remove that asset to the list of selected assets (the first video demonstrate it). Do you think it is worth it to add it to the carousel preview too?

I don't think I'd add it to the carousel/preview.

etoledom:
Thanks @iamthomasbishop for the input!

@frosty Let's close this PR as it is, and I'll open a new PR to tackle the audio files as Thomas described. What do you think?

Pod/Classes/WPCarouselAssetsViewController.m
```diff
@@ -0,0 +1,135 @@
...
+- (void)setPreviewingAssetAtIndex:(NSInteger)index animated:(BOOL)animated
+{
+    self.index = index;
+    if (self.isViewLoaded) {
+        UIViewController *newViewController = [self viewControllerAtIndex:index];
+        [self setViewController:newViewController animated:animated];
+    }
+}
+
+#pragma mark: - UIPageViewControllerDelegate
```
frosty: `pragma mark`s don't need a `:` before the `-` – just `#pragma mark - UIPageViewControllerDelegate`. There are a couple in here that need updating.

Pod/Classes/WPCarouselAssetsViewController.m
```diff
@@ -0,0 +1,135 @@
...
+    id<WPMediaAsset> asset = assetController.asset;
+    return [self.assets indexOfObject:asset];
+}
+
+- (UIViewController *)viewControllerAtIndex:(NSInteger)index
+{
+    id<WPMediaAsset> asset = [self.assets objectAtIndex:index];
+    WPAssetViewController *fullScreenImageVC = [[WPAssetViewController alloc] init];
+    fullScreenImageVC.asset = asset;
+    //    fullScreenImageVC.selected = [self positionOfAssetInSelection:asset] != NSNotFound;
```
frosty: Can this line be removed?

etoledom:
Thank you @frosty !

PR #282, opened 2018-04-04
Bottom action bar

etoledom:
This PR implements a bottom action bar to preview and insert assets selected under multiple selection.

The [original design](https://user-images.githubusercontent.com/2722505/37941996-994ad1ea-31a3-11e8-9660-19badea57e28.jpg) was made for Stock Photos, and the original issue can be found [here](https://github.com/wordpress-mobile/WordPress-iOS/issues/8954).

- Added an option to continue using the top right button in case of any library user prefers that.
- Small changes in the use of `self.internalSelectedAssets` and `self.selectedAssets` to avoid unnecessary copy.

bottom_bar

**To test:**

- Open the media picker with multiple selection.
- Play selecting, deselecting, previewing, inserting assets, showing and dismissing the keyboard.

### DIFF
--- a/Example/WPMediaPicker/AppDelegate.m
+++ b/Example/WPMediaPicker/AppDelegate.m
@@ -13,6 +13,9 @@
     // Customize appearance
     self.window.tintColor = [UIColor colorWithRed:0/255.0f green:135/255.0f blue:190/255.0f alpha:1.0f];
     self.window.backgroundColor = [UIColor whiteColor];
+
+    //Configure bottom action bar line color
+    [[WPActionBar appearance] setLineColor:[UIColor colorWithRed:0.91 green:0.94 blue:0.95 alpha:1.0]];
     
     //Configure navigation bar background color
     [[UINavigationBar appearance] setBarTintColor:[UIColor colorWithRed:0/255.0f green:135/255.0f blue:190/255.0f alpha:1.0f]];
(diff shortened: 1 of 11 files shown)


etoledom:
Thanks @frosty for the review!

**iPhone X**:
I couldn't make the UIToolBar work nicely with the iPhone X bottom safe area. The tool bar seems to have a completely fixed height, and adding a constraint (as I saw some people recommending) breaks the nice dismiss animation and the keyboard support, and it just looks weird.

Instead, I made a custom view to use as the action bar. With a few constraints tricks it works perfectly with safe areas, and it seems to work just fine with scroll view insets too.

We also win in customizability, and we won't have the colors issue in the main app.

I added to it a top line that is not visible by default (prioritizing defaults for WPiOS) but I set a custom color to it in the demo app to make it a bit prettier.

screen shot 2018-04-05 at 5 09 25 pm

>We won't want to show the bar in the standalone media library

Setting `options.showActionBar = false` will solve that issue 👍

etoledom:
Thanks a lot @frosty for the review!

>on iPad, when the picker is displayed in a popover it looks a little odd having the bar at the bottom of the screen.

I noticed that too and I agree. The accessory input view is set on a different window, so to fix that we will most probably have to set it manually as a subview of the controller's view.

Anyway, I tested that in the main app and all the popover pickers (that I found) are single selection, so for now that's not yet a problem I guess.

iamthomasbishop:
@frosty Agreed that it looks odd, and attached to the bottom of the popover would probably be the best place for the action bar. 👍

PR #288, opened 2018-04-24
Enable bounce correctly depending of scrolling type.

SergioEstevao:
When using horizontal scroll on full screen the collection view has allowing vertical bounce, this PR fix that.

To test:
 - Open the demo app
 - Open options
 - Disable vertical scrolling
 - Open the picker full screen
 - Select an album
 - Scroll horizontally and make sure no bounce happens vertically


frosty:
Hmm, I thought the vertical scroll option was only supposed to affect the input picker, not fullscreen?

frosty:
To clarify, I think it should have no effect on the fullscreen picker – that should always only scroll horizontally.

SergioEstevao:
I don't think it should be limited to the input picker, we are not using it but it can be used both ways.

PR #295, opened 2018-05-30
Add badge to 

etoledom:
This PR fixes part of [issue #9353](https://github.com/wordpress-mobile/WordPress-iOS/issues/9353), adding support for a custom badge at the top right corner of the picker cell.

The badge will show the file type.

## Configuration:

To configure what types show, a new option `badgedUTTypes` was introduced, with a list of file types that we want to show it's type in the badges.

## WPMediaAsset Conformance:

This also uses a new optional `WPMediaAsset` method `UTTypeIdentifier`, to get the file type information.

We could use the file extension, but using UTTypes seems to ensure the correct identification of the file.

We can also return `nil` if we don't want to show the badge of a filetype under specific conditions (like a gif with no animations)

## PHAsset issue:

The implementation of `UTTypeIdentifier` in the `PHAsset` was a bit tricky:
Using `[[PHAssetResource assetResourcesForAsset:self] firstObject] uniformTypeIdentifier];` was very slow, to the point that scrolling the picker gets choppy.

I tested using `[self valueForKey:@"uniformTypeIdentifier"];` and the performance is quite good. The downside of this is that it looks like it's using private api. I saw many people on StackOverflow using this, so we might not get rejected, but I'm not sure if we should take the risk.

## Style:
I opted to use the system blur effect, and adding a touch with vibrancy effect.
@iamthomasbishop could you please take a look?
PD: There is a small shadow but it's not visible in the screenshot for some reason :( 

img_2861

To test:
- Save some GIFs images in your device image library.
- Run the app and see the badge appearing.


Pod/Classes/WPBadgeView.m
```diff
@@ -0,0 +1,141 @@
...
+
+- (void)style
+{
+    self.label.font = [UIFont systemFontOfSize:14.f weight:UIFontWeightSemibold];
+    self.label.textColor = UIColor.whiteColor;
+    self.insets = kDefaultEdgeInsets;
+    self.cornerRadius = kDefaultCornerRadius;
+}
+
+- (void)dropShadow
```
SergioEstevao: Maybe call this `setupDropShadow`, dropShadow only looks like a property...

Pod/Classes/WPBadgeView.m
```diff
@@ -0,0 +1,141 @@
...
+{
+    _cornerRadius = cornerRadius;
+    self.blurEffectView.layer.cornerRadius = cornerRadius;
+    self.blurEffectView.layer.masksToBounds = YES;
+    [self setNeedsDisplay];
+}
+
+#pragma mark - Helpers
+
+- (void)style
```
SergioEstevao: Rename to setupStyle?

Pod/Classes/WPMediaPickerOptions.h
```diff
@@ -43,4 +43,9 @@
  */
 @property (nonatomic, assign) BOOL showActionBar;
 
+/**
+ A list of UTTypes where the picker cell should show a badge showing the file type. (i.e. UTTypeGif)
+ */
+@property (nonatomic, strong) NSArray<NSString *> *badgedUTTypes;
```
SergioEstevao: What about using a NSSet instead of a NSArray?

Pod/Classes/WPPHAssetDataSource.m
```diff
@@ -558,6 +558,11 @@ - (NSString *)fileExtension
     return [[[[PHAssetResource assetResourcesForAsset:self] firstObject] originalFilename] pathExtension];
 }
 
+- (nullable NSString *)UTTypeIdentifier
+{
+    return [self valueForKey:@"uniformTypeIdentifier"];
```
SergioEstevao: I'm ok with using the private setter, but if this property changes name, will this method just return nil?

Pod/Classes/WPBadgeView.m
```diff
@@ -0,0 +1,141 @@
...
+- (instancetype)init
+{
+    self = [super initWithFrame:(CGRectZero)];
+    if (self) {
+        [self initialSetup];
+    }
+    return self;
+}
+
+- (void)initialSetup
```
SergioEstevao: Just to be consistent with other view classes on the pod, can we rename this method to `commonInit` ?

Pod/Classes/WPPHAssetDataSource.m
etoledom: Right! If the key where to change/be removed, we would get a awful crash: 
`... this class is not key value coding-compliant for the key ...`
We could:
- Take the risk.
- Let the scroll get slower with the safer method.
- Not show the badge on device media picker 😞 

I'll try to find another option for this, but so far, I haven't found a safe&fast way.

jklausa: you can do `valueForKey:@selector(uniformTypeIdentifier)` and if you rename it you'll get a compiler error

jklausa: (I think)

etoledom: Cool! Thank you!
Using that we can also check for `respondsToSelector`. That would save us from the runtime crashes in production.

etoledom:
Thanks @SergioEstevao !

Added the requested changes. And also thanks @jklausa for pointing me to the right direction to make that `valueForKey` call safer.

Now, if the key were to be removed/changed, we will get a compile warning, and it won't crash at runtime.

And also, as requested by @iamthomasbishop , I removed the vibrancy effect. Now it shows the plane white text in the badge:

img_2862

etoledom:
Thank you @SergioEstevao and @bummytime !

frosty:
@etoledom Just a note – while it probably won't have much effect, we could slightly improve performance of the shadows by providing a `shadowPath`. I can push a separate PR for that if you like.

etoledom:
Thanks @frosty ! If you can do it would be great.
But I can send it too if you are busy.

iamthomasbishop:
Looking more closely, the final badge design didn't have a shadow (but the original did). They don't bother me, but if it's any better/more performant, we can remove the shadow 😄

frosty:
@iamthomasbishop just to clarify, would you prefer if it did or didn’t have a shadow?

etoledom:
@frosty We just had a conversation with @iamthomasbishop , and we decided to remove it all together. I'm preparing the PR for that now :)

PR #296, opened 2018-05-31
Remove shadow from WPBadgeView

etoledom:
From the comments [here](https://github.com/wordpress-mobile/MediaPicker-iOS/pull/295#issuecomment-393595737), we decided to remove the shadow effect from `WPBadgeView`.

This PR does just that.
I decided to not keep the shadow as a optional setup for sake of simplicity.

To test:
- Make sure you have GIF images in your device media library.
- Run the app and check that the GIF badge renders correctly (and without a shadow)

### DIFF
--- a/Pod/Classes/WPBadgeView.m
+++ b/Pod/Classes/WPBadgeView.m
@@ -1,8 +1,6 @@
 
 #import "WPBadgeView.h"
 
-static const CGFloat kDefaultShadowRadius = 2.f;
-static const CGFloat kDefaultShadowOpacity = 0.5f;
 static const CGFloat kDefaultCornerRadius = 6.f;
 static const UIEdgeInsets kDefaultEdgeInsets = {3.f, 6.f, 3.f, 6.f};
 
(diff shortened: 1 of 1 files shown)


Pod/Classes/WPBadgeView.m
```diff
@@ -79,15 +76,13 @@ - (void)setInsets:(UIEdgeInsets)insets
     self.bottomConstraint.constant = -insets.bottom;
     self.leadingConstraint.constant = insets.left;
     self.trailingConstraint.constant = -insets.right;
-    [self setNeedsLayout];
```
frosty: Are you sure this isn't necessary any more? Or perhaps `[self layoutIfNeeded]`?

etoledom: I tested it with:
```
    cell.badgeView.cornerRadius = 24;
    cell.badgeView.insets = UIEdgeInsetsMake(15, 15, 15, 15);
```
From `cellForItemAtIndexPath` and it worked. So it seems that it's not necessary.

img_0003

But you are right, in other use cases it might be necessary so I will add the `[self layoutIfNeeded]` call.
 
Thanks for pointing it out!

PR #303, opened 2018-07-31
Adding support to use a View Controller for the empty view.

ScoutHarris:
Ref #https://github.com/wordpress-mobile/WordPress-iOS/issues/9504

This adds the ability to use a `ViewController` for the `emptyView` state, instead of just a `UIView`. Specifically because we are transitioning from `WPNoResultsView` to `NoResultsViewController` in `WordPress-iOS` .

To test:
- This can be fully testing with https://github.com/wordpress-mobile/WordPress-iOS/pull/9889, which replaces the NoResultsView for the Stock Photo Picker.
- Otherwise, just a sanity check/code review.

Hey @etoledom - it looks like you might be familiar with the `emptyView` logic. Would you mind taking a look at this? Thanks!!

### DIFF
--- a/Pod/Classes/WPMediaPickerViewController.h
+++ b/Pod/Classes/WPMediaPickerViewController.h
@@ -253,12 +253,6 @@
 @property (nonatomic, strong, readonly, nonnull) UILabel *defaultEmptyView;
 
 /**
- The default empty view controller. When `emptyViewControllerForMediaPickerController:` is not implemented,
- use this property to style the message.
- */
-@property (nonatomic, strong, readonly, nonnull) UIViewController *defaultEmptyViewController;
-
-/**
  A localized string that reflect the action that will be done when the user finishes picking assets.
  This string can contain a a placeholder for a numeric value that will indicate the number of media items selected.
  If this is nil the default value will be used. The default the value is 'Add %@'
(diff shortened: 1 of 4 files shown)


Pod/Classes/WPMediaPickerViewController.h
```diff
@@ -234,10 +247,17 @@
...
+ The default empty view. When `emptyViewForMediaPickerController:` is not implemented,
+ use this property to style the message.
  */
 @property (nonatomic, strong, readonly, nonnull) UILabel *defaultEmptyView;
 
+/**
+ The default empty view controller. When `emptyViewControllerForMediaPickerController:` is not implemented,
+ use this property to style the message.
+ */
+@property (nonatomic, strong, readonly, nonnull) UIViewController *defaultEmptyViewController;
```
etoledom: Is it necessary to have both a `defaultEmptyView` and a `defaultEmptyViewController`?
Probably we should just have one default that will be displayed if non of the related delegates method are implemented. Does it make sense?

Not sure which would be the best option, but I'd vote for the simplest of them.

ScoutHarris: Hey @etoledom . Good point...

So, `WPNavigationMediaPickerViewController` needs access to `defaultEmptyViewController`, so I added a method definition. I also removed `defaultEmptyViewController` as a property, and changed it to use `defaultEmptyView`.

ScoutHarris:
Hey @etoledom . Thanks for the timely review and helpful feedback!

Some comments in response - 

> It reappears under the keyboard.

Adding `[self centerEmptyView];` did in fact fix it.


> When the search is cancelled, it looks like the empty controller disappear and then appears again in the lower position. Is it possible to make it slide down the same way it slide up when the keyboard appears?

It was not being re-added, it was just re-centering _immediately_. So I added a `animateWithDuration` to slow it down a bit.

Otherwise, it's ready for another review!


> It also seems to disappear when the search bar is cleared.

I'm not seeing this. If you still do, can you provide more information? (device, iOS version, anything else helpful)

PR #304, opened 2018-08-08
Adding check for emptyViewController.

ScoutHarris:
Ref #https://github.com/wordpress-mobile/WordPress-iOS/issues/9940

Since the Post Editor in WPiOS sometimes return nil for `emptyViewController`, it was crashing when attempting to add it to the view. This adds a test to ensure it actually exists before doing so.

To test:
- This can be tested with https://github.com/wordpress-mobile/WordPress-iOS/pull/9943 by verifying edit/create a post does not crash.

### DIFF
--- a/Pod/Classes/WPMediaPickerViewController.m
+++ b/Pod/Classes/WPMediaPickerViewController.m
@@ -599,7 +599,7 @@ static CGFloat SelectAnimationTime = 0.2;
 
 - (void)addEmptyViewControllerToView
 {
-    if (self.emptyViewController.view.superview == nil) {
+    if (self.emptyViewController && _emptyViewController.view.superview == nil) {
         [self.collectionView addSubview:self.emptyViewController.view];
         _emptyViewController.view.frame = self.collectionView.frame;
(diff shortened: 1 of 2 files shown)


ScoutHarris:
Hey @frosty . Would you mind doing this one with https://github.com/wordpress-mobile/WordPress-iOS/pull/9943 since they go together? Thank you!!

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -599,7 +599,7 @@ - (UILabel *)defaultEmptyView
 
 - (void)addEmptyViewControllerToView
 {
-    if (self.emptyViewController.view.superview == nil) {
+    if (self.emptyViewController && _emptyViewController.view.superview == nil) {
```
frosty: This method switches between `self.emptyViewController` and `_emptyViewController` a few times – could we make them all the same?

ScoutHarris: Will do. Thanks @frosty !

PR #309, opened 2018-09-19
Adjust placement of empty view while searching

ScoutHarris:
Ref #https://github.com/wordpress-mobile/WordPress-iOS/issues/10178

With the above WPiOS change, a search with no results displays a different empty view. This adjusts the placement of the empty view if a search is being performed.

To test:
- This can be fully testing with https://github.com/wordpress-mobile/WordPress-iOS/pull/10179, which updates all the search no results views.
- Otherwise, just a sanity check/code review.


ScoutHarris:
Hey @jklausa . Would you mind doing this with https://github.com/wordpress-mobile/WordPress-iOS/pull/10179? They are linked, and this is a small change. Thank you!!

PR #312, opened 2019-01-09
Attempt to fix crash in performBatchUpdates

yaelirub:
In performBatchUpdates we are updating a local variable that will also be returned from numberOfItems delegate method in an attempt
to return the right number of items after an update (insert or delete rows) has occurred.

Fixes [#10073](https://github.com/wordpress-mobile/WordPress-iOS/issues/10073) (maybe)

To test:
I was not able to reproduce this crash. To test:
Go to Media and add/delete items. 
Make sure it works as expected. (items are added/ deleted and not crash).

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - WPMediaPicker (1.3.1-beta.1)
+  - WPMediaPicker (1.3.1)
 
 DEPENDENCIES:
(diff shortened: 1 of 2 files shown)


yaelirub:
@etoledom thank you so much for looking into this. I will test again and fix

yaelirub:
Hey @etoledom , @jkmassel Im going to close this one and try again.

PR #320, opened 2019-04-15
Version 1.3.3 with fixed warning

jtreanor:
This is a minor change to fix a nullability warning in `WPMediaPickerViewController.h`. Doing this so we have one less warning in https://github.com/wordpress-mobile/WordPress-iOS/pull/11381.


jtreanor:
Thanks, Jeremy. Yea, I think it's fine to keep these together since the changes are all simple.

Issue #305, opened 2018-08-28
Crash Occurred when opening picker


StephanWalters:
Crash Occurred when opening the picker. It's only happened once, but providing the crashlog:
http://crashes.to/s/c8e25eec50a

##### Tested on [device], iOS [version], WPMediaPicker [version]
- iPhone 
- iOS 11
- WPMediaPicker 1.3

SergioEstevao:
Solved on release 1.3.4

PR #321, opened 2019-04-23
Issue/10073 crash on batch changes

SergioEstevao:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/10073 #305 

This PR makes to changes to the codebase:
 - Makes sure we send a NO to the change observer block when no incrementalChanges are detected.
 - Make sure we removed the previous observers when changing dataSources.
 
To test:
 - Open the demo app
 - Press on the + icon
 - Check if photos shows correctly
 - Switch to the Photos App, and add new assets to it or remove some
 - Switch to the Demo app and see if updates happen correctly
 - Test on the device and simulator

 @frosty Can you test with your massive iCloud library? :)

### DIFF
--- a/Pod/Classes/WPPHAssetDataSource.m
+++ b/Pod/Classes/WPPHAssetDataSource.m
@@ -75,7 +75,7 @@
 
         if (!groupChangeDetails && !assetsChangeDetails && !albumChangeDetails) {
             [self.observers enumerateKeysAndObjectsUsingBlock:^(NSUUID *key, WPMediaChangesBlock block, BOOL *stop) {
-                block(true, [NSIndexSet new], [NSIndexSet new], [NSIndexSet new], @[]);
+                block(false, [NSIndexSet new], [NSIndexSet new], [NSIndexSet new], @[]);
             }];
             return;
         }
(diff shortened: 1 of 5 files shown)


Pod/Classes/WPPHAssetDataSource.m
```diff
@@ -75,7 +75,7 @@ - (void)photoLibraryDidChange:(PHChange *)changeInstance
 
         if (!groupChangeDetails && !assetsChangeDetails && !albumChangeDetails) {
             [self.observers enumerateKeysAndObjectsUsingBlock:^(NSUUID *key, WPMediaChangesBlock block, BOOL *stop) {
```
yaelirub: Im curious, why are we enumerating on the observers if there were no changes? or am I misunderstanding the code?

yaelirub:
Followed test steps and it seems to be working as expected.

Pod/Classes/WPPHAssetDataSource.m
SergioEstevao: Because when there is no details of the changes but the delegate was called is because you need to do a full reload. So there is still changes there but delegate is not able to send the details of it.

PR #333, opened 2019-06-14
Fix preview frame media sizes

SergioEstevao:
The asset preview was not being correctly centered on phones with no home button.

This PR updates the screen to take in account the top and bottom guidelines.

To test:
 - Start the demo app
 - Tap on +
 - Open the camera roll
 - LongPress or ForceTouch an asset to see it's preview
 - Check that image/video is centered.
 - Rotate the device and check that media is still centered.

### DIFF
--- a/Pod/Classes/WPAssetViewController.m
+++ b/Pod/Classes/WPAssetViewController.m
@@ -36,15 +36,15 @@
     self.imageView.translatesAutoresizingMaskIntoConstraints = NO;
     [self.imageView.leftAnchor constraintEqualToAnchor:self.view.leftAnchor].active = YES;
     [self.imageView.widthAnchor constraintEqualToAnchor:self.view.widthAnchor].active = YES;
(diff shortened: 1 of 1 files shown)


etoledom:
Hey @SergioEstevao 

I've been testing and comparing with develop, and I don't really see any difference 🤔 
Is there a special size/aspect-ratio of images/videos that makes this issue show up?

SergioEstevao:
@etoledom did you test on iPhone X kind of phone? I think the main thing here is to take in account the bottom space used by the home indicator.

frosty:
I just tested this on the XR, and the image appears to not be centered in the black space – there's more space below than above. It _is_ centered in the entire screen space though (if you measure from the very top of the screen to the very bottom, including the navigation bar). Is that expected?

SergioEstevao:
@etoledom and @frosty I updated the code, it should now be centered in the center of the available (black) screen.

PR #343, opened 2020-03-10
Update MediaPicker-iOS pod

### LINKED ISSUE #342: Update MediaPicker-iOS pod to Swift 5
In the podspec, define Swift version 5: `s.swift_version = '5.0'`. This project is using only Objective-C, so there isn't a flag in the project file to define the Swift version at this time.

mindgraffiti:
Fixes #342.

This PR:
-  updates the schema to the latest available in Xcode 11.3.1
-  defines a Swift version in the podspec

The project is written in Objective-C, so there is no project variable to define as Swift 5 at this time.

**To test:**
1. Check out this branch
2. Build and run


mindgraffiti:
Whoops. Added reviewers too soon. Investigating the CircleCI failures.

SergioEstevao:
I think we will need to bump the pod version number to 1.6.1 just to make sure the changes are not picked by users of the previous version.

mindgraffiti:
@SergioEstevao that's true. I updated the podspec version, but the CircleCI tests are failing before it even hits the Podspec validation. Something in CircleCI's `bundle install` fails to install JSON 2.2.0 using ruby.

SergioEstevao:
@mindgraffiti if you merge from develop the CI issue should be solved now.

WPMediaPicker.podspec
```diff
@@ -16,7 +16,8 @@ Pod::Spec.new do |s|
   s.source           = { :git => "https://github.com/wordpress-mobile/MediaPicker-iOS.git", :tag => s.version.to_s }
 
   s.platform     = :ios, '11.0'
-  s.requires_arc = true 
+  s.requires_arc = true
```
bjtitus: Wonder if this can just be removed since the default is `true`
https://guides.cocoapods.org/syntax/podspec.html#requires_arc

mindgraffiti: That's interesting! TIL. I didn't add this parameter, it already existed. I think I fixed the whitespace on it. What do you think about adding a task that reviews all of our .podspecs and makes sure they're all consistent?

mindgraffiti:
@Gio2018 yeah good call. @bjtitus since I'm removing more warnings from the project, I'll remove that extra property from the podspec too (`s.requires_arc = true`)

@osullivanchris could you make more app icons and a launch image? I'm rusty and forget what it takes to make these. Here's a list of the missing ones:

Screen Shot 2020-03-11 at 1 47 40 PM

Screen Shot 2020-03-11 at 1 47 52 PM

Screen Shot 2020-03-11 at 1 48 03 PM

mindgraffiti:
@osullivanchris I also updated the launch image and checked it in a variety of size formats. This removes the launch image warning in Xcode.

**iPad Pro 12.9 inch **
Simulator Screen Shot - iPad Pro (12 9-inch) - 2020-03-11 at 15 37 32

** iPhone SE **


** iPhone 11 Pro Max **

WPMediaPicker.podspec
bjtitus: > What do you think about adding a task that reviews all of our .podspecs and makes sure they're all consistent?

👍 That could make sense. At least 10 dependencies seem to have the flag and I did notice some inconsistency with the Readmes which might be nice to unify just to make it easier to identify the current version.

mindgraffiti:
Thank you @Gio2018 @bjtitus 🙇‍♀

Issue #349, opened 2020-04-14
NSRangeException in RCTModalHostViewController


loremattei:
**Sentry Url**: https://sentry.io/share/issue/fe06bc567c1f41a6852b71769221f4d4/
**User Count**: 350
**Count**: 855
**First Release**: 13.4.1.1
**First Seen**: 2019-11-05T22:49:50Z
**Last Release**: 14.5.0.2
**Last Seen**: 2020-04-14T09:55:01Z
**24 Hours**: 7
**30 Days**: 180

**Stacktrace**:
```
NSRangeException: *** -[NSMutableIndexSet addIndexesInRange:]: Range {18446744073709551613, 1} exceeds maximum index value of NSNotFound - 1
    at null.__exceptionPreprocess(null:null)
    at null.objc_exception_throw(null:null)
    at null.-[NSMutableIndexSet addIndexesInRange:](null:null)
    at null.__NSINDEXSET_IS_CALLING_OUT_TO_A_BLOCK__(null:null)
    at null.__NSIndexSetEnumerate(null:null)
    at null.-[WPPHAssetDataSource adjustedIndexesForIndexSet:forCount:](WPMediaPicker/Pod/Classes/WPPHAssetDataSource.m:364)
    at null.-[WPPHAssetDataSource adjustedIndexesForIndexSet:](WPMediaPicker/Pod/Classes/WPPHAssetDataSource.m:358)
    at null.__45-[WPPHAssetDataSource photoLibraryDidChange:]_block_invoke(WPMediaPicker/Pod/Classes/WPPHAssetDataSource.m:102)
    at null._dispatch_call_block_and_release(null:null)
    at null._dispatch_client_callout(null:null)
...
(8 additional frame(s) were not displayed)
```

leandroalonso:
@frosty or @SergioEstevao can you please give me some insight into this one?

This crash happens when the photo library change (`photoLibraryDidChange`). One of the returned indexes is probably a `NSNotFound`, which causes the crash.

I thought that something like this might fix the issue:

```
- (NSIndexSet *)adjustedIndexesForIndexSet:(NSIndexSet *)indexes forCount:(NSInteger)count
{
    NSMutableIndexSet *adjustedSet = [NSMutableIndexSet new];
    [indexes enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL * _Nonnull stop) {
        if (idx != NSNotFound) {
[...]

PR #352, opened 2020-04-27
Check NSNotFound before adjusting indexes

leandroalonso:
Fixes https://github.com/wordpress-mobile/MediaPicker-iOS/issues/349

### To test

I wasn't able to reproduce the original issue. Just check if the Media Picker behaves correctly.


leandroalonso:
@SergioEstevao I've seen here that you have an upcoming PR. Should I release a new MediaPicker version and integrate with WPiOS or you're planning to do it after merging your PR?

PR #353, opened 2020-05-20
Make sure we don't interfere on show after VC navigation bars.

SergioEstevao:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/9685

To test:
 - Start the demo app
 - Select the Options on Top
 - Activate the Show Post Processing Step
 - Press Done
 - Press on + on the Top Right
 - Select an asset
 - Check that the empty post-processing step shows and navigation bar is ok.

 - Test the bug above using this PR:

### DIFF
--- a/Pod/Classes/WPNavigationMediaPickerViewController.m
+++ b/Pod/Classes/WPNavigationMediaPickerViewController.m
@@ -13,6 +13,7 @@ UIPopoverPresentationControllerDelegate
 @property (nonatomic, strong) WPMediaPickerViewController *mediaPicker;
 @property (nonatomic, strong) WPMediaGroupPickerViewController *groupViewController;
 @property (nonatomic, strong) NSObject *changesObserver;
+@property (nonatomic, weak) UIViewController *afterSelectionViewController;
 @end
 
 @implementation WPNavigationMediaPickerViewController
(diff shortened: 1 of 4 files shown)


Pod/Classes/WPNavigationMediaPickerViewController.m
```diff
@@ -382,6 +386,7 @@ - (void)navigationController:(UINavigationController *)navigationController will
 - (void)showAfterViewController:(UIViewController *)viewController
 {
     NSParameterAssert(viewController);
+    self.afterSelectionViewController = viewController;
```
pinarol, line 389: It is not very clear to me what an `afterSelectionViewController` is, can you explain a bit?

pinarol:
Tested this with WPiOS app by changint he Podfile as:

> pod 'WPMediaPicker', :git => 'https://github.com/wordpress-mobile/MediaPicker-iOS.git', :commit => '6313e850dd6f47c40fdc5b6e6fec010c54d727a0'

Seems to solve the issue.

Pod/Classes/WPNavigationMediaPickerViewController.m
SergioEstevao: So this method was something that was added to WPMediaPicker to cater to the scenarios that you want to show a VC in the same navigation controller of the picker after a selection is made. This was added to cater for the Crop option that we see in the main apps. 

Before this was available we were closing the picker and open another screen in modal, but the UI felt very bad, so we implemented this.

I agree with you that this name is not great, do you have a suggestion for a better name?

pinarol: OK thanks, I don't have any problem with naming, I was just trying to understand how this is solving. 👍

PR #355, opened 2020-07-06
Update project for Xcode 12 and iOS 14

SergioEstevao:
This PR updates the code for Xcode 12 and iOS 14:

-  Check for PhotoLibrary permissions correctly in iOS 14.

On iOS 14 there is a new permission system for accessing the media library. 

The method to check for permission was deprecated and a new method was added: `requestAuthorizationForAccessLevel`

The answers to the new method have a new option called "PHAuthorizationStatusLimited" and it allows users to limit the media that is available to third-parties.

After a user selects this new option, the app will allow changing the media selected, after a "cold" restart.

This PR updates the code to use the new method when available. 

I'm considering if we should alert the user that they restricted their media access so only a small subset is available and direct the user to the settings to change that.

To test:
 - Start the demo app
 - Press the + button
 - Check that access to Photos permissions are requested and the option to "Select Media" works correctly.


SergioEstevao:
@frosty and @loremattei do you mind give a quick look at this one?

loremattei (approved):
This looks good to me, but I'm not super familiar with media, so another pair of eyes would be appreciated.

Pod/Classes/WPPHAssetDataSource.m
```diff
@@ -119,61 +119,70 @@ - (void)photoLibraryDidChange:(PHChange *)changeInstance
...
-                        return;
-                    }
-                    case (WPMediaLoadOptionsAssets): {
-                        [self loadAssetsWithSuccess:successBlock failure:failureBlock];
-                        return;
+                return;
+            }
+            case PHAuthorizationStatusNotDetermined:
+            {
+                [PHPhotoLibrary requestAuthorization:^(PHAuthorizationStatus status) {
```
frosty: I may not be following the code correctly, but should we be doing the iOS 14 check here again, or not? How come we don't call `requestAuthorizationForAccessLevel` again here? 🤔

frosty (approved):
This looks good to me, and testing works as expected – I just had one small query about the code, but I suspect I'm just not following something :)

Pod/Classes/WPPHAssetDataSource.m
SergioEstevao: You are absolutely correct! Just updated the code.

frosty (approved):
Looks good, thanks for changing :)

PR #360, opened 2020-09-24
Allow MediaPickerViewControllerDelegate errors to be nullable

frosty:
This PR allows the error parameter passed to `MediaPickerViewControllerDelegate.mediaPickerController(_:handleError:)` to be null. This seems to be one of the causes of https://github.com/wordpress-mobile/WordPress-iOS/issues/14958 in my investigation, where we call a failure block (due to no assets being present), but also have no error to pass.

**To test**

- Build and and run the [associated WPiOS branch](https://github.com/wordpress-mobile/WordPress-iOS/pull/14989) on iOS 14
- Visit the Media Library, choose +, tap Choose From My Device, then Select More Photos (or Select Photos if this is the first time you've run the app). Ensure that no photos are selected and tap Done.
- The media picker should be displayed with no content, and the app should not crash.

### DIFF
--- a/Pod/Classes/WPNavigationMediaPickerViewController.m
+++ b/Pod/Classes/WPNavigationMediaPickerViewController.m
@@ -351,7 +351,7 @@ static NSString *const ArrowDown = @"\u25be";
     [self updateSelectionAction];
 }
 
-- (BOOL)mediaPickerController:(nonnull WPMediaPickerViewController *)picker handleError:(nonnull NSError *)error {
+- (BOOL)mediaPickerController:(nonnull WPMediaPickerViewController *)picker handleError:(nullable NSError *)error {
     if ([self.delegate respondsToSelector:@selector(mediaPickerController:handleError:)]) {
         return [self.delegate mediaPickerController:picker handleError:error];        
     } else {
--- a/Pod/Classes/WPMediaPickerViewController.h
+++ b/Pod/Classes/WPMediaPickerViewController.h
@@ -218,7 +218,7 @@
  *  @param error The error to show
  *  @return YES if the error was handled by the delegate
  */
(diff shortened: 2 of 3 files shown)


Pod/Classes/WPNavigationMediaPickerViewController.m
```diff
@@ -351,7 +351,7 @@ - (void)mediaPickerController:(nonnull WPMediaPickerViewController *)picker sele
     [self updateSelectionAction];
 }
 
-- (BOOL)mediaPickerController:(nonnull WPMediaPickerViewController *)picker handleError:(nonnull NSError *)error {
+- (BOOL)mediaPickerController:(nonnull WPMediaPickerViewController *)picker handleError:(nullable NSError *)error {
```
diegoreymendez, line 354: It's weird that we're calling `mediaPickedController:handleError:` with a `nil` error.  Is there a valid use-case for allowing this scenario?

IMHO it would be far cleaner if we used this logic instead:

1. If there's no error: avoid this whole chain of methods from being called.
2. If there's an error: make sure `error` is properly instantiated.
3. Or if the first two are too tricky to implement, at least make sure we bail out here if error is `nil` without calling the method in the delegate.

Just my 2 cents.

diegoreymendez (commented):
Added a comment, since I feel the approach may be allowing for a big chain of calls resulting in a no-op.

Pod/Classes/WPNavigationMediaPickerViewController.m
frosty: Following up from our Slack discussion, good point that it doesn't necessarily make sense to call `handleError` with no error. I'll push a PR shortly with a different approach to fix this crash.

frosty:
Closed, preferring #361.

PR #367, opened 2021-08-11
Add Buildkite

jkmassel:
Adds Buildkite support to the project, including:
- Fixing the ruby security issue
- Adds fastlane
- Adds rubocop
- Updates Cocoapods
- Fixes issues identified by rubocop

**To Test:**
- Note that Buildkite CI passes
- Note that [Pod Publishing works](https://buildkite.com/wordpress-mobile/mediapicker-ios/builds/14#5f3fd05f-2085-4424-9f70-1c8ab434dcec)


AliSoftware (approved):
Code changes make sense and BuildKite CI passes, so approving.

The CircleCI "Test" job fails, apparently because it couldn't find the right device/simulator installed on the CircleCI VM image.

As a side note, I also noticed that [the test used to compare Xcode versions](https://github.com/wordpress-mobile/circleci-orbs/blob/trunk/src/ios/orb.yml#L317) in this step is incorrect, as it compare the versions alphanumerically (String comparison) instead of using `Gem::Version` comparison (like we do [here](https://github.com/wordpress-mobile/circleci-orbs/blob/trunk/src/ios/orb.yml#L280)). Obviously since we're migrating away from CircleCI and this orb, this is not worth fixing anymore, but just wanted to mention it in case a similar logic was reproduced in one of the BuildKite scripts we wrote while we ported the orb, commands and pipelines from CircleCI to BuildKite, and could have copied that mistake over there.

Maybe, since that CircleCI check currently fails anyway, we could consider disabling it in `.circleci/config.yml` already with this PR before merging it – given the test step was ported to BuildKite and seems to work already, contrary to its CircleCI's counterpart… wdyt @jkmassel ?

jkmassel:
Good call – this repo isn't too busy, so migrating it fully to Buildkite should be fine. 

There's no Xcode version check logic anywhere in the Buildkite stuff, so we should be safe.

Thanks for the thorough review!

PR #368, opened 2021-08-26
Update project to be  iOS 13 minimum version

SergioEstevao:
This PR updates the podspec to request iOS 13 as a minimum.
It also updates the code to reflect that change and removes any deprecation warnings.

One of the most relevant deprecation was the removal of support of ForceTouch  preview and adopt long press for all devices.

To test:
 - Start the demo app
 - Open the media picker
 - Long Press on asset and check that preview works correctly


Example/Podfile.lock
```diff
@@ -1,5 +1,5 @@
 PODS:
-  - WPMediaPicker (1.7.2)
```
frosty, line 2: I'm not sure why the pod version has changed to 0.99.9? The latest on `develop` seems to be 1.7.3. I guess if we're following some kind of semver this should perhaps become 2.0.0 (as it breaks backwards compatibility)?

Pod/Classes/WPMediaCapturePreviewCollectionView.m
```diff
@@ -110,7 +110,8 @@ - (void)startCapture
                         CALayer *viewLayer = self.previewView.layer;
                         self.captureVideoPreviewLayer.frame = viewLayer.bounds;
                         self.captureVideoPreviewLayer.videoGravity = AVLayerVideoGravityResizeAspectFill;
-                        self.captureVideoPreviewLayer.connection.videoOrientation = [self videoOrientationForInterfaceOrientation:[[UIApplication sharedApplication] statusBarOrientation]];
+                        UIWindowScene *currentScene = [[[[UIApplication sharedApplication] windows] lastObject] windowScene];
```
frosty, line 113: Now I think about it, we could keep the current iOS version requirement and just make this conditional? If iOS 13 use windowScene, otherwise use the previous code. What do you think?

Example/Podfile.lock
SergioEstevao: There was a mistake made in develop some time ago that made the pod version to be 0.99.9, It's was fixed but because I didn't update here to latest develop we had this remaining.
It's now fixed.

Pod/Classes/WPMediaCapturePreviewCollectionView.m
SergioEstevao: There was some more deprecation errors and updates that I did in this project to make it work correctly in iOS 13 and forward. 
I don't think it makes sense for us to keeping supporting older version of iOS and have our code with conditionals.

SergioEstevao: What do you say?

frosty: I think that's fine, I just wondered if there was any benefit retaining some compatibility but I think it does make sense to update.

frosty (approved):

PR #372, opened 2021-09-22
371 Check conformance on call to optional method

### LINKED ISSUE #371: NSInvalidArgumentException: optional property accessed in -[WPMediaCollectionViewCell configureAccessibility] without checking for conformance
**Describe the bug**
[A low-incidence crash in the WooCommerce app](https://github.com/woocommerce/woocommerce-ios/issues/3722) was found, caused by an `NSInvalidArgumentException -[WooCommerce.CancellableMedia filename]: unrecognized selector sent to instance`. 

`filename` an `@optional` property on the `WPMediaAsset` [Objective-C protocol definition](https://github.com/wordpress-mobile/MediaPicker-iOS/blob/45726d0ba935f16846efca45d320f72ea4c74b2a/Pod/Classes/WPMediaCollectionDataSource.h#L169), along with `fileExtension` and `UTTypeIdentifier`.

This was found to be called from the `-[WPMediaCollectionViewCell configureAccessibility]` method, for documents, without first checking for `respondsToSelector:`, causing a crash where users have assets other than images, video, or audio, in their Media Library.

Unable to reproduce at present – there is likely a very uncommon corner case leading to this crash in the WooCommerce app, because all the uses of MediaPicker-iOS in the app are currently filtered to show images only.

To reproduce, I would suggest invoking the picker without a filetype filter, using an asset implementation which does not include `filename`, on a site with a document. It probably requires a demo app to repro reliably.

**Expected behavior**
The app should not crash when a document is present in the media library, even if `filename` is unimplemented on the asset implementation.

joshheald:
Fixes #371 

## Description
A rare crash found in WooCommerce iOS was caused by access to the `filename` of an asset, which is an optional member of the `WPMediaAsset` protocol. In WCiOS, this was not implemented, so the app crashed with `NSInvalidArgumentException -[WooCommerce.CancellableMedia filename]: unrecognized selector sent to instance`

To prevent this, we check for `respondsToSelector:` before getting the filename, and default to using the creation date in the accessibility labels for documents when filename is not implemented.

The fallback to creation date matches the existing approach taken for other media types.

## To test
Reproducing the crash in WCiOS is challenging, as it's in a path which _shouldn't_ be followed by WCiOS, because the API request for media items is filtered to images only, and this crash occurs when there are assets shown which are of type `WPMediaTypeOther`.

Some small code changes in the WooCommerce app make it much easier to reproduce. 
1. To see the crashing behaviour, use this branch: [`hack/example-media-picker-crash`](https://github.com/woocommerce/woocommerce-ios/tree/hack/example-media-picker-crash)
2. Connect to a site with a non-audio/video/image document in the media library.
3. Open a media picker, e.g. Products > Add > Add Product Image
4. If the document is in the first page of results, the app will crash, otherwise scroll until it shows to see the crash.

To test with the fix
1. Check out this branch of the media picker, and update the WCiOS Podfile to include `pod 'WPMediaPicker', :path => "path/to/MediaPicker-iOS"`
2. Run `bundle exec rake dependencies`
3. Follow steps above, app will not crash.
4. Check VoiceOver for the document, it will be announced as "Document: creation date" (using the fallback because filename is not implemented)

Crash | Fix
------|------
Crash | Fixed

### DIFF
--- a/Pod/Classes/WPMediaCollectionViewCell.m
+++ b/Pod/Classes/WPMediaCollectionViewCell.m
@@ -210,7 +210,11 @@ static const CGFloat LabelRegularFontSize = 13;
             accessibilityLabel = [NSString stringWithFormat:NSLocalizedString(@"Audio, %@", @"Accessibility label for audio items in the media collection view. The parameter is the creation date of the audio."), formattedDate];
             break;
         case WPMediaTypeOther:
-            accessibilityLabel = [NSString stringWithFormat:NSLocalizedString(@"Document: %@", @"Accessibility label for other media items in the media collection view. The parameter is the filename file."), [_asset filename]];
+            if([_asset respondsToSelector:@selector(filename)]) {
+                accessibilityLabel = [NSString stringWithFormat:NSLocalizedString(@"Document: %@", @"Accessibility label for other media items in the media collection view. The parameter is the filename file."), [_asset filename]];
+            } else {
+                accessibilityLabel = [NSString stringWithFormat:NSLocalizedString(@"Document, %@", @"Accessibility label for other media items in the media collection view. The parameter is the creation date of the document."), formattedDate];
+            }
             break;
         default:
             break;
--- a/WPMediaPicker.podspec
+++ b/WPMediaPicker.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WPMediaPicker'
-  s.version       = '1.7.2'
+  s.version       = '1.7.3-beta.1'
 
   s.summary       = 'WPMediaPicker is an iOS controller that allows capture and picking of media assets.'
   s.description   = <<-DESC


SergioEstevao (approved):
Great catch! Thanks for fixing this!

joshheald:
@SergioEstevao thanks for the review! I've just gone to follow the next step in [how to do a pod release](https://iosp2.wordpress.com/2019/10/08/shipping-a-pod-release/) (5: publish the new version) and I've found that [1.7.2 was created but never released](https://github.com/wordpress-mobile/MediaPicker-iOS/releases/tag/1.7.2).

Do you think I should continue to add a new 1.7.3-beta release regardless? Or put this stuff in to 1.7.2? Presumably we don't really want to skip a version number.

SergioEstevao:
If there is not tag for 1.7.2 and WPiOS doesn't refer to it (by a commit) feel free to add to it.

PR #376, opened 2022-01-07
PHPhotoLibrary: Register change observer after asking for permission

### LINKED ISSUE #375: Xcode 13.2.1 - Photos not loading in iOS 15.2 simulator
### Problem
In the WPMediaPicker-Example app, while trying to access the photos from a simulator running iOS 15.2 I faced weird freezes I wasn’t able to access the photos.

This happens when building the app using Xcode 13.2.1

### Steps
1. There was no response for a few seconds after tapping on `Tap here to quick select assets"
2. I see the alert asking for permission access ONLY after a few seconds has passed.
3. Tapping on Allow Access to All Photos dismissing the permission alert.
4. I don’t see any photos.

I have tested in a simulator running iOS 14.1 and things are working as expected there.

### Screen recording
https://user-images.githubusercontent.com/524475/148616808-fd7816ab-1dd1-46b3-a8be-c5752b4c9c2a.mov

selanthiraiyan:
Fixes #375 

### Description
As proposed in https://developer.apple.com/forums/thread/696804 this PR register's `PHPhotoLibrary` change observer after asking for permission from the user.

### Steps to test
- Using Xcode 13.2.1 - Run the example app in a device/simulator running iOS 15.2 and make sure that you can access photos without any issues.
- Also, make sure that the changes done to the photos library reflect in the app. (To ensure that the observer is working as expected)

Note: I have only tested the above points in a simulator and it would be great if you can test this in a device.

### DIFF
--- a/Pod/Classes/WPPHAssetDataSource.m
+++ b/Pod/Classes/WPPHAssetDataSource.m
@@ -47,7 +47,6 @@
     _refreshGroups = YES;
     _cachedCollections = [[NSMutableArray alloc] init];
     _imageGenerationQueue = dispatch_queue_create("org.wordpress.wpmediapicker.WPPHAssetDataSource", DISPATCH_QUEUE_SERIAL);
-    [[PHPhotoLibrary sharedPhotoLibrary] registerChangeObserver:self];
     return self;
 }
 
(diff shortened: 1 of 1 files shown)


SergioEstevao (approved):
Great fix! Thanks for this!

SergioEstevao:
@jkmassel I've tested this fix and it solves a critical crash on WPMediaPicker for iOS version 15.2 

After this PR is merged we should create a new version of the pod and update all the projects that use this pod.

From what I see the issue identified only happens when the apps are installed for the first and it request permissions to access the device Media library. So this may not affect existing users, but can/will affect new users of the apps.

PR #380, opened 2022-01-20
Invalidate custom header layout if trait collection changes

frosty:
This PR invalides the layout for a custom header (if provided) in the case that the trait collection changes. This allows the custom header view to potentially resize itself to handle the trait change (for example, when the dynamic type size changes).

**To test**

Please test using the steps in the associated WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/17795


sla8c (approved):
LGTM!

PR #383, opened 2022-03-15
Add permission check when adding media

sla8c:
This PR adds a permission check + functionality to feedback to the user that a photo captured from camera could not be added to support WordPress-iOS issue https://github.com/wordpress-mobile/WordPress-iOS/issues/17467

Ref #https://github.com/wordpress-mobile/WordPress-iOS/issues/17467

To test

Please test using the steps in the associated WPiOS PR: wordpress-mobile/WordPress-iOS#18139

### DIFF
--- a/Pod/Classes/WPMediaPickerHelper.h
+++ b/Pod/Classes/WPMediaPickerHelper.h
@@ -1,0 +1,7 @@
+#import <Foundation/Foundation.h>
+
+@interface WPMediaPickerHelper : NSObject
+
++ (UIAlertController *)buildAlertControllerWithError:(NSError *)error okActionHandler:(void (^ __nullable)(UIAlertAction *action))handler;
+
+@end
--- a/Pod/Classes/WPPHAssetDataSource.m
+++ b/Pod/Classes/WPPHAssetDataSource.m
@@ -131,11 +131,7 @@
 {
     [self checkPermissionStatus:^(PHAuthorizationStatus status) {
 
-        /// Starting from iOS 15.2 we should do the registration
-        /// after asking user for permission
-        /// Solution proposed here - https://developer.apple.com/forums/thread/696804
-        ///
-        [[PHPhotoLibrary sharedPhotoLibrary] registerChangeObserver:self];
+        [self registerPHChangeObserver];
 
         switch (status) {
             case PHAuthorizationStatusRestricted:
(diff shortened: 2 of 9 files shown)


Pod/Classes/WPMediaPickerHelper.h
```diff
@@ -0,0 +1,7 @@
+#import <Foundation/Foundation.h>
+
+@interface WPMediaPickerHelper : NSObject
```
frosty: Great to pull this code out, but I wonder if this class could have a name more closely related to what it does? `Helper` is very generic 🙂

Pod/Classes/WPPHAssetDataSource.m
```diff
@@ -450,40 +450,86 @@ - (void)addAssetWithChangeRequest:(PHAssetChangeRequest *(^)(void))changeRequest
...
-    } completionHandler:^(BOOL success, NSError *error) {
-        if (!success) {
-            if (completionBlock){
-                dispatch_async(dispatch_get_main_queue(), ^{
-                    completionBlock(nil, error);
-                });
+
+    [self checkPermissionStatus:^(PHAuthorizationStatus status) {
+
+        /// Starting from iOS 15.2 we should do the registration
```
frosty: There's a lot of similarity here between this implementation in `addAssetWithChangeRequest:` and `loadDataWithOptions:success:failure:`. Could we do anything to consolidate the two or share code at all? We probably also don't need this exact comment duplicated.

I wonder if we could move the call to `registerChangeObserver` into `checkPermissionStatus` and call it before we call the handler block, seeing as we now need it in two places?

frosty (commented):
Thanks for picking this up! This is working for me as described in the testing steps 👍, but I've left a few thoughts about tweaking the code structure.

Pod/Classes/WPMediaPickerHelper.h
sla8c: That's done! 🙂

Pod/Classes/WPPHAssetDataSource.m
sla8c: I've created a function for handling `registerChangeObserver` + will drop you a message to discuss the other(s). Thanks!

frosty (approved):
Thanks for making those changes! :shipit:!

frosty (approved):
Changes look good, except the pod spec version can just be 1.8.3-beta.1, as that's the next increment from the released version on `trunk`.

PR #394, opened 2022-09-06
Fix incremental update

crazytonyli:
## Issue
When the media library is updated incrementally _with changed items_, the collection view shows incorrect thumbnails for some items.

This issue can be easily reproduced from [WordPress-iOS](https://github.com/wordpress-mobile/WordPress-iOS/) repo trunk branch. To reproduce this issue,
1. Change ContextManager in WordPress-iOS repo to [using `ContainerContextFactory`](https://github.com/wordpress-mobile/WordPress-iOS/blob/7f44120b43ca36401791036912e834ade49f79c9/WordPress/Classes/Utility/ContextManager.m#L53).
2. Launch WordPress app from Xcode.
3. [Desktop browser] Go to a WordPress.com site.
4. [Desktop browser] Add or delete photos in the site's media library.
5. [WordPress app] Go to the site's "Media" screen in the launched WordPress-iOS app.
6. Repeat step 4 and 5, check out the collection view in the media library screen.

https://user-images.githubusercontent.com/1101828/188596443-ff1efbf6-104b-4ec1-b949-5933dbfc13c6.mov

## Change

During batch updates, the changed items were reloaded manually by calling the configuring cell method, instead of using the collection view. I believe this approach causes the issue since index path of the item in collection view doesn't necessary match the index path of the media library data, not _during_ batch updates.

This PR moves the reloading changed items code to the batch update's completion block, which seems like a more reliable place to reload items.


CHANGELOG.md
```diff
@@ -3,34 +3,39 @@ All notable changes to this project will be documented in this file.
...
+- `0.24` Release  - [0.24](#024)
+- `0.23` Release  - [0.23](#023)
+- `0.22` Release  - [0.22](#022)
+- `0.21` Release  - [0.21](#021)
+- `0.20` Release  - [0.20](#020)
+- `0.19` Release  - [0.19](#019)
+- `0.18` Releases - [0.18](#018)
+- `0.17` Releases - [0.17](#017)
+- `0.16` Releases - [0.16](#016)
+- `0.15` Releases - [0.15](#015)
```
mokagio: For what is worth, I think we could get rid of this table of contents.

Having it is not a pattern I've seen elsewhere in the repo. The internal links look suspiciously like PRs.

mokagio (approved):
I'm pretty rusty when it comes to `UICollectionView`, but I think this change makes sense.

From the [docs](https://developer.apple.com/documentation/uikit/uicollectionview/1618045-performbatchupdates), it doesn't seem like accessing indexes within the `update` block is recommended, so moving that code afterwards seems like the way to go.

> Deletes are processed before inserts in batch operations. This means the indexes for the deletions are processed relative to the indexes of the collection view’s state before the batch operation, and the indexes for the insertions are processed relative to the indexes of the state after all the deletions in the batch operation.

mokagio:
@crazytonyli what do you think about updating this in the clients?

It seems to me like an edge case bug, so not worth targeting the frozen branch?

crazytonyli:
@mokagio Looking at the doc, it seems it's more appropriate to reload the items with other updates all together. I've updated this PR accordingly, which now only changed how changed item are reloaded and seems like a minimal change.

The app isn't affected by this issue actually since the app is using `LegacyContextFactory`, not `ContainerContextFactory`, and this issue only occurs when `ContainerContextFactory` is used.

CHANGELOG.md
crazytonyli: Deleted in https://github.com/wordpress-mobile/MediaPicker-iOS/pull/394/commits/61afeb219f4f37392234d59a3001588201bbefa5 😄

mokagio:
> @mokagio Looking at the doc, it seems it's more appropriate to reload the items with other updates all together. I've updated this PR accordingly, which now only changed how changed item are reloaded and seems like a minimal change.

Well there you go. You always learn something new, even from the docs you (_me in this case!_) linked yourself.

> The app isn't affected by this issue actually since the app is using LegacyContextFactory, not ContainerContextFactory, and this issue only occurs when ContainerContextFactory is used.

Cool. No rush in upgrading then.

Re-iterating my approval. Merge and ship a new version at will.

PR #395, opened 2022-10-04
Tools Update: `bash-cache` Buildkite plugin version 2.8.0, Xcode 14

mokagio:
This `bash-cache` version includes a fix that will remove the need for the `gem install bundler` workarounds.


.buildkite/pipeline.yml
```diff
@@ -1,10 +1,10 @@
 # Nodes with values to reuse in the pipeline.
 common_params:
   plugins: &common_plugins
-  - &bash_cache automattic/bash-cache#2.7.0: ~
+  - &bash_cache automattic/bash-cache#2.8.0
```
spencertransier, line 4: @mokagio What was the `: ~` doing? I haven't seen that syntax before.

spencertransier (approved):
Just had a question about syntax I hadn't seen before. These changes look great! :shipit:

.buildkite/pipeline.yml
mokagio: The Buildkite docs use this syntax with plugins that require no arguments, as can be seen with a search for `~` in [this docs page](https://buildkite.com/docs/plugins/using).

However, everything works without this, so I've been removing it whenever I've seen it.

PR #396, opened 2022-10-19
Add `CHANGELOG.md`

mokagio:
Adds `CHANGELOG.md`. Based on https://github.com/wordpress-mobile/WordPressKit-iOS/pull/545 and the PR template iteration from https://github.com/wordpress-mobile/WordPressUI-iOS/pull/119.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,7 +1,7 @@
 # Nodes with values to reuse in the pipeline.
 common_params:
   plugins: &common_plugins
-  - &bash_cache automattic/bash-cache#2.8.0
+  - &bash_cache automattic/bash-cache#2.9.0
   # Common environment values to use with the `env` key.
   env: &common_env
     IMAGE_ID: xcode-14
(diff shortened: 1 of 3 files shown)


AliSoftware:
@mokagio CI is failing due to that default bundle issue. You'll probably need to update the `bash-cache-buildkite-plugin` while you're at it (like you did on the similar PRs of some of the other repos).

crazytonyli:
Will there be any issue if the old change logs are kept? Is it cause some kind of parsing issue?

mokagio:
@crazytonyli ouch! I didn't notice that 🤦‍♂️ 

I'll revert the change so that the previous changes are maintained.

As for whether the different format will result in parsing issues... maybe? We don't have an official or portable parser at the moment. The closest thing is [the one used in `release-toolkit`](https://github.com/wordpress-mobile/release-toolkit/blob/2abbaa3faae263e660dc3b237a62c14aed06e292/rakelib/changelog_parser.rake). I think that one would not work because it expect "Trunk" where this has "Unreleased".

I guess we'll address this scenario once we'll get to automate extracting the release notes in all the pods?

crazytonyli:
@mokagio All good. I thought they were deleted intentionally 😸

crazytonyli (approved):

PR #399, opened 2022-10-26
Fix position of empty view to avoid message disappearing when keyboard shows up

### LINKED ISSUE #18840: Media library: 'No media matching your search' message is no visible
### Expected behavior
While searching for media in Media Library, `No media matching your search` message is displayed when no results are found.
Also, if the search is cancelled and the user tries to add a media file again, they will see all available media in the search screen.

### Actual behavior
When searching Media Library without results, `No media matching your search` message is not visible, but underneath the header.
Also, when the search is canceled by tapping the `Cancel` header button the `No media matching your search` message will be displayed if the user tries to add Media Library items again for any media block.

### Steps to reproduce the behavior
1. Create a new blog post
2. Add a Image block
3. Click `ADD IMAGE`
4. Click `WordPress Media Library`
5. Search for some string that will not return any results
6. ❌ `No media matching your search` message is not visible. _Expected: The message IS visible._
7. Click `Cancel` button in Navigation bar
8. Click `ADD IMAGE`
9. Click `WordPress Media Library`
10. ❌ `No media matching your search` message is visible. _Expected: Available images are displayed_

https://user-images.githubusercontent.com/42008628/172656786-618ba8ee-b2da-4bd2-bc8c-d6cc3c686f0d.MP4

diogo-lins:
Fixes [#18840](https://github.com/wordpress-mobile/WordPress-iOS/issues/18840), [#19514](https://github.com/wordpress-mobile/WordPress-iOS/issues/19514) 

## Description

- Changing the `emptyViewController` to manipulate your size and position using **constraints** to the View, instead of using frame manipulation.
- Changing the bottom constraint of the `emptyViewController` to follow the **Keyboard height** when the same appears.

### Issue

https://user-images.githubusercontent.com/20734091/197835979-4dbc38c4-3cc8-4ac3-8c08-69f6f1b60c03.mp4

### After changes

| Blog Post - With Images | Blog Post - Without Images | Media - With Images | Media - Without Images |
| ------------------------ | ------------------------ | ------------------------ | ------------------------|
| Blog-WithImages | Blog-WithoutImages | Media-WithImages | Media-WithoutImages |

-----
| Gray space before | Gray space after |
| ------------------- | ------------------- |
|  |  |

## Testing instructions

This can be tested by cloning my fork(and branch) and changing the `WordPress-iOS` Podfile to use the Pod locally:
```
pod 'WPMediaPicker', :path => '...'
```

1. Run the app
2. Switch to a site that has no media uploaded
3. Open Blog Post and add an `IMAGE` block
4. Make sure that the custom empty view is displayed correctly
5. Opens the Keyboard
6. Make sure that the message `No media matching your search` is displayed correctly

### DIFF
--- a/Pod/Classes/WPMediaPickerViewController.h
+++ b/Pod/Classes/WPMediaPickerViewController.h
@@ -222,6 +222,9 @@
  *  @return An empty view to display or `nil` to not display any.
  *
  *  If this method is not implemented, a default UILabel will be displayed.
+ *
+ *  In case of (emptyViewControllerForMediaPickerController) and (emptyViewForMediaPickerController) are both implemented by a parent ViewController,
+ *  the first one has precedence over the second one to be presented to the user.
  */
 - (nullable UIView *)emptyViewForMediaPickerController:(nonnull WPMediaPickerViewController *)picker;
 
@@ -235,6 +238,9 @@
  *
  *  If this method is not implemented, a default ViewController with a default
  *  UILabel will be displayed.
+ *
+ *  In case of (emptyViewControllerForMediaPickerController) and (emptyViewForMediaPickerController) are both implemented by a parent ViewController,
+ *  the first one has precedence over the second one to be presented to the user.
  */
 - (nullable UIViewController *)emptyViewControllerForMediaPickerController:(nonnull WPMediaPickerViewController *)picker;
 
--- a/Pod/Classes/WPMediaPickerViewController.h
+++ b/Pod/Classes/WPMediaPickerViewController.h
@@ -223,8 +223,7 @@
  *
  *  If this method is not implemented, a default UILabel will be displayed.
  *
- *  In case of (emptyViewControllerForMediaPickerController) and (emptyViewForMediaPickerController) are both implemented by a parent ViewController,
- *  the first one has precedence over the second one to be presented to the user.
+ * `emptyViewControllerForMediaPickerController` takes precedence over this method.
  */
 - (nullable UIView *)emptyViewForMediaPickerController:(nonnull WPMediaPickerViewController *)picker;
 
@@ -239,8 +238,7 @@
  *  If this method is not implemented, a default ViewController with a default
  *  UILabel will be displayed.
  *
- *  In case of (emptyViewControllerForMediaPickerController) and (emptyViewForMediaPickerController) are both implemented by a parent ViewController,
- *  the first one has precedence over the second one to be presented to the user.
+ *  This method takes precedence over `emptyViewForMediaPickerController`.
  */
 - (nullable UIViewController *)emptyViewControllerForMediaPickerController:(nonnull WPMediaPickerViewController *)picker;
 
--- a/Pod/Classes/WPMediaPickerViewController.m
+++ b/Pod/Classes/WPMediaPickerViewController.m
@@ -51,7 +51,7 @@ static NSString *const CustomHeaderReuseIdentifier = @"CustomHeaderReuseIdentifi
 @property (nonatomic, strong) UILabel *defaultEmptyView;
 @property (nonatomic, strong) UIViewController *emptyViewController;
 @property (nonatomic, strong) UIViewController *defaultEmptyViewController;
-
+@property (nonatomic, strong) NSLayoutConstraint *emptyViewBottomConstraint;
 
 @property (nonatomic, strong) WPActionBar *accessoryActionBar;
 @property (nonatomic, strong) UIButton *selectedActionButton;
@@ -97,15 +97,13 @@ static CGFloat SelectAnimationTime = 0.2;
 {
     [super viewDidLoad];
 
-    self.refreshControl = [[UIRefreshControl alloc] init];
-    [self.refreshControl addTarget:self action:@selector(pullToRefresh:) forControlEvents:UIControlEventValueChanged];
-    [self.collectionView addSubview:self.refreshControl];
-
     // Setup subviews
+    [self setupPullToRefresh];
     [self addCollectionViewToView];
-    [self addEmptyViewToView];
     [self setupCollectionView];
     [self setupSearchBar];
+    [self addEmptyViewToView];
+
     [self setupLayout];
 
     //setup data
@@ -117,6 +115,13 @@ static CGFloat SelectAnimationTime = 0.2;
     self.layout.sectionInsetReference = UICollectionViewFlowLayoutSectionInsetFromSafeArea;
     
     [self refreshDataAnimated:NO];
+}
+
+- (void)setupPullToRefresh
+{
+    self.refreshControl = [[UIRefreshControl alloc] init];
+    [self.refreshControl addTarget:self action:@selector(pullToRefresh:) forControlEvents:UIControlEventValueChanged];
+    [self.collectionView addSubview:self.refreshControl];
 }
 
 - (void)registerDataSourceObservers {
@@ -623,13 +628,26 @@ static CGFloat SelectAnimationTime = 0.2;
 
 - (void)addEmptyViewControllerToView
 {
-    if (self.emptyViewController && self.emptyViewController.view.superview == nil) {
-        [self.collectionView addSubview:self.emptyViewController.view];
-        self.emptyViewController.view.frame = self.collectionView.frame;
-        [self addChildViewController:self.emptyViewController];
-        [self.emptyViewController didMoveToParentViewController:self];
-        [self centerEmptyView];
+    if (!self.emptyViewController && self.emptyViewController.view.superview != nil) {
+        return;
     }
+
+    [self addChildViewController:self.emptyViewController];
+    [self.emptyViewController.view setTranslatesAutoresizingMaskIntoConstraints:NO];
+    [self.collectionView addSubview:self.emptyViewController.view];
+
+    self.emptyViewBottomConstraint = [self.emptyViewController.view.bottomAnchor constraintEqualToAnchor:self.view.bottomAnchor];
+
+    [NSLayoutConstraint activateConstraints:
+     @[
+       [self.emptyViewController.view.topAnchor constraintEqualToAnchor:self.collectionView.topAnchor],
+       self.emptyViewBottomConstraint,
+       [self.emptyViewController.view.leadingAnchor constraintEqualToAnchor:self.view.leadingAnchor],
+       [self.emptyViewController.view.trailingAnchor constraintEqualToAnchor:self.view.trailingAnchor]
+       ]
+     ];
+
+    [self.emptyViewController didMoveToParentViewController:self];
 }
 
 - (void)removeEmptyViewControllerFromView
@@ -1419,6 +1437,7 @@ referenceSizeForFooterInSection:(NSInteger)section
     }
     self.collectionView.contentInset = contentInset;
     self.collectionView.scrollIndicatorInsets = contentInset;
+    [self.emptyViewBottomConstraint setConstant:-keyboardFrameEnd.size.height];
 
     [UIView animateWithDuration:0.2 animations:^{
         [self centerEmptyView];
@@ -1432,6 +1451,7 @@ referenceSizeForFooterInSection:(NSInteger)section
     contentInset.bottom = 0.f;
     self.collectionView.contentInset = contentInset;
     self.collectionView.scrollIndicatorInsets = contentInset;
+    [self.emptyViewBottomConstraint setConstant:0.f];
 
     [UIView animateWithDuration:0.2 animations:^{
         [self centerEmptyView];
@@ -1445,35 +1465,18 @@ referenceSizeForFooterInSection:(NSInteger)section
 - (void)centerEmptyView
 {
     if (self.emptyViewController) {
-        CGRect emptyViewFrame = [self getEmptyViewFrame];
-
-        if ([self.searchBar.text isEqualToString:@""]) {
-            emptyViewFrame.origin.y -= self.searchBar.frame.size.height/2;
-        }
-        emptyViewFrame.size.height -= self.view.layoutMargins.bottom;
-        emptyViewFrame.size.height -= self.view.layoutMargins.top;
-        _emptyViewController.view.frame = emptyViewFrame;
+        [self.view layoutIfNeeded];
     } else {
         self.emptyView.center = self.collectionView.center;
         self.emptyView.frame = [self getEmptyViewFrame];
-    }
+   }
 }
 
 - (CGRect)getEmptyViewFrame
 {
-    CGRect emptyViewFrame;
+    CGRect emptyViewFrame = self.emptyView.frame;
 
-    if (_emptyViewController) {
-        emptyViewFrame = self.collectionView.frame;
-    } else {
-        emptyViewFrame = self.emptyView.frame;
-    }
-
-    CGFloat superviewHeight = self.collectionView.frame.size.height;
-    CGFloat totalInsets = self.collectionView.contentInset.top + self.collectionView.contentInset.bottom;
-
-    superviewHeight = superviewHeight - totalInsets > 0 ? superviewHeight - totalInsets : superviewHeight;
-    emptyViewFrame.origin.y = (superviewHeight / 2.0) - (emptyViewFrame.size.height / 2.0) + self.collectionView.frame.origin.y;
+    emptyViewFrame.origin.y = self.collectionView.frame.origin.y;
 
     return emptyViewFrame;
 }
--- a/Pod/Classes/WPMediaPickerViewController.m
+++ b/Pod/Classes/WPMediaPickerViewController.m
@@ -46,6 +46,7 @@ static NSString *const CustomHeaderReuseIdentifier = @"CustomHeaderReuseIdentifi
 
 @property (nonatomic, strong, readwrite) UISearchBar *searchBar;
 @property (nonatomic, strong) NSLayoutConstraint *searchBarTopConstraint;
+@property (nonatomic, assign) CGFloat currentKeyboardHeight;
 
 @property (nonatomic, strong) UIView *emptyView;
 @property (nonatomic, strong) UILabel *defaultEmptyView;
@@ -637,6 +638,7 @@ static CGFloat SelectAnimationTime = 0.2;
     [self.collectionView addSubview:self.emptyViewController.view];
 
     self.emptyViewBottomConstraint = [self.emptyViewController.view.bottomAnchor constraintEqualToAnchor:self.view.bottomAnchor];
+    [self.emptyViewBottomConstraint setConstant:-self.currentKeyboardHeight];
 
     [NSLayoutConstraint activateConstraints:
      @[
@@ -1441,7 +1443,8 @@ referenceSizeForFooterInSection:(NSInteger)section
     }
     self.collectionView.contentInset = contentInset;
     self.collectionView.scrollIndicatorInsets = contentInset;
-    [self.emptyViewBottomConstraint setConstant:-keyboardFrameEnd.size.height];
+    self.currentKeyboardHeight = keyboardFrameEnd.size.height;
+    [self.emptyViewBottomConstraint setConstant:-self.currentKeyboardHeight];
 
     [UIView animateWithDuration:0.2 animations:^{
         [self centerEmptyView];
@@ -1455,7 +1458,8 @@ referenceSizeForFooterInSection:(NSInteger)section
     contentInset.bottom = 0.f;
     self.collectionView.contentInset = contentInset;
     self.collectionView.scrollIndicatorInsets = contentInset;
-    [self.emptyViewBottomConstraint setConstant:0.f];
+    self.currentKeyboardHeight = 0.f;
+    [self.emptyViewBottomConstraint setConstant:self.currentKeyboardHeight];
 
     [UIView animateWithDuration:0.2 animations:^{
         [self centerEmptyView];
--- a/Pod/Classes/WPMediaPickerViewController.m
+++ b/Pod/Classes/WPMediaPickerViewController.m
@@ -593,6 +593,7 @@ static CGFloat SelectAnimationTime = 0.2;
 - (void)addContainerEmptyView
 {
     if (self.emptyViewContainer != nil && self.emptyViewContainer.superview != nil) {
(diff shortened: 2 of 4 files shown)


WPMediaPicker.podspec
```diff
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WPMediaPicker'
-  s.version       = '1.8.5'
+  s.version       = '1.8.6'
```
hassaanelgarem: Please update the podspec version to be a beta.
```suggestion
  s.version       = '1.8.6-beta.1'
```

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -1445,35 +1465,18 @@ - (void)keyboardWillHideNotification:(NSNotification *)notification
...
-
-    if (_emptyViewController) {
-        emptyViewFrame = self.collectionView.frame;
-    } else {
-        emptyViewFrame = self.emptyView.frame;
-    }
-
-    CGFloat superviewHeight = self.collectionView.frame.size.height;
-    CGFloat totalInsets = self.collectionView.contentInset.top + self.collectionView.contentInset.bottom;
+    CGRect emptyViewFrame = self.emptyView.frame;
```
hassaanelgarem: I see we still do some frame manipulation when the delegate doesn't provide an empty view controller and we fall back on using `emptyView`.

What do you think of adding a "container" view? If we go that route, we'll add constraints to the container view and hide/show it as needed. And we'll embed in it, either the view provided by the delegate or the default view. This way we unify the logic between the two possible paths. WDYT?

hassaanelgarem (commented):
The grey space issue is fixed ✅ 

I found one issue with the no search results string placement. In landscape, the string is partially hidden behind the keyboard on small-screen devices.
The below screenshot is from an iPhone SE:

hassaanelgarem:
I've also encountered another lower-priority issue.

1. Open a site with a single media asset
2. Go to My Site -> Menu -> Media
3. Select the existing asset and delete it
4. Try searching for anything
5. Delete the search query
6. Notice that the "You don't have any media" screen is hidden behind the keyboard.

WPMediaPicker.podspec
diogo-lins: Got it! I'll do it!

diogo-lins:
> The grey space issue is fixed ✅
> 
> I found one issue with the no search results string placement. In landscape, the string is partially hidden behind the keyboard on small-screen devices. The below screenshot is from an iPhone SE:
> 
> 

Yeah, you're right, after investigation, this "bug" was created after the fix for this PR, I believe this fix should be made on https://github.com/wordpress-mobile/WordPress-iOS, we need to update the constraint on the `NoResultsViewController` for this case, this VC that manipulates the text top constraint, should I create an issue and raise the new PR there after this PR be merged and bump the pod version on the WordPress project?

I also mention this issue here: p1666286767668469-slack-C011BKNU1V5

_Actual behavior:_
The below screenshot is from my personal iPhone 13:

unnamed

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: I think is feasible to do it 😄, I decided to keep this doesn't have a side effect in another area, and to not change something that is already done on the code, but let me prepare some code with this approach in the next commit. Thanks for this!

diogo-lins:
> I've also encountered another lower-priority issue.
> 
> 1. Open a site with a single media asset
> 2. Go to My Site -> Menu -> Media
> 3. Select the existing asset and delete it
> 4. Try searching for anything
> 5. Delete the search query
> 6. Notice that the "You don't have any media" screen is hidden behind the keyboard.
> 
> 

Got it, we just need to update the bottom constraint in this situation too, let me do this. 😄 

Resolved here https://github.com/wordpress-mobile/MediaPicker-iOS/pull/399/commits/887de53fbad039df81ba3eb869123248ba6f3d72 😄 

centerEmptyVC

diogo-lins:
Hey @hassaanelgarem 👋🏾, how's it going? I have a question: If a Parent View Controller is using the media picker and implements these two delegates:
```
- (nullable UIView *)emptyViewForMediaPickerController:(nonnull WPMediaPickerViewController *)picker;
- (nullable UIViewController *)emptyViewControllerForMediaPickerController:(nonnull WPMediaPickerViewController *)picker;
```
Who has more priority to show? the View Controller or the UIView?

WPMediaPicker.podspec
diogo-lins: Resolved here https://github.com/wordpress-mobile/MediaPicker-iOS/pull/399/commits/1c26b25530e77de5731f00471173961302855e79 😄

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: Change made here https://github.com/wordpress-mobile/MediaPicker-iOS/pull/399/commits/ded72ca809b30d35e75f2c6213e4be8951586531 😄 

emptyView

hassaanelgarem:
> Who has more priority to show? the View Controller or the UIView?

I'm not sure, tbh. What is the current behavior?
I'd say, whatever the current behavior is, let's keep it. And let's make the priority clear in the documentation of each function.

diogo-lins:
> > Who has more priority to show? the View Controller or the UIView?
> 
> I'm not sure, tbh. What is the current behavior? I'd say, whatever the current behavior is, let's keep it. And let's make the priority clear in the documentation of each function.

Following the current behavior, the empty VC has more priority.

WPMediaPicker.podspec
hassaanelgarem: Given that `1.8.6` has [recently been released](https://github.com/wordpress-mobile/MediaPicker-iOS/commit/8f261918313dd551cc7394d78d0574b458064579), I think this should now be `1.8.7-beta.1`

hassaanelgarem:
The single media asset issue is resolved 🚀 

> we need to update the constraint on the NoResultsViewController for this case, this VC that manipulates the text top constraint, should I create an issue and raise the new PR there after this PR be merged and bump the pod version on the WordPress project?

There's no need to block this fix until #399 is merged, we should actually aim to test both concurrently.

Please create a PR on WPiOS with this fix before we merge this PR. For now, the WPiOS PR should point to the `WPMediaPicker` living on the branch `issue/18840-19514-emptyview-height-v2`. Once #399 is merged and the new version is released, we can update the WPiOS PR to point to the released version.

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -602,14 +631,14 @@ - (UIView *)emptyView
     return _emptyView;
 }
 
-- (void)addEmptyViewToView
+/** Checks if the parentViewController is providing a custom empty ViewController to be added, if not, add a provided custom emptyView
+ */
+- (void)selectEmptyView
```
hassaanelgarem: The naming of this function feels a bit off. At first glance, it seems like the user has selected something. I understand that naming is tricky since we have four functions that do very similar things. But what about the following names:
- `addEmptyViewContainer`
- `populateEmptyViewContainer`
- `addEmptyViewToContainer`
- `addEmptyViewControllerToContainer`

What do you think?

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -46,8 +46,10 @@ @interface WPMediaPickerViewController ()
 
 @property (nonatomic, strong, readwrite) UISearchBar *searchBar;
 @property (nonatomic, strong) NSLayoutConstraint *searchBarTopConstraint;
+@property (nonatomic, assign) CGFloat currentKeyboardHeight;
 
 @property (nonatomic, strong) UIView *emptyView;
+@property (nonatomic, strong) UIView *containerEmptyView;
```
hassaanelgarem: I think this is more verbose
```suggestion
@property (nonatomic, strong) UIView *emptyViewContainer;
```

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -624,37 +653,68 @@ - (UILabel *)defaultEmptyView
...
+       [self.emptyViewController.view.bottomAnchor constraintEqualToAnchor:self.containerEmptyView.bottomAnchor],
+       [self.emptyViewController.view.leadingAnchor constraintEqualToAnchor:self.containerEmptyView.leadingAnchor],
+       [self.emptyViewController.view.trailingAnchor constraintEqualToAnchor:self.containerEmptyView.trailingAnchor]
        ]
      ];
 
     [self.emptyViewController didMoveToParentViewController:self];
 }
 
+- (void)removeContainerEmptyView
```
hassaanelgarem: Can he hide/show the container instead of removing and adding it? It seems more convenient and will simplify the code. But I'm not sure it has drawbacks. What do you think?

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -1455,34 +1519,21 @@ - (void)keyboardWillHideNotification:(NSNotification *)notification
     contentInset.bottom = 0.f;
     self.collectionView.contentInset = contentInset;
     self.collectionView.scrollIndicatorInsets = contentInset;
-    [self.emptyViewBottomConstraint setConstant:0.f];
+    self.currentKeyboardHeight = 0.f;
+    [self.emptyViewBottomConstraint setConstant:self.currentKeyboardHeight];
```
hassaanelgarem: [Nit] Just for the sake of consistency 😄 
```suggestion
    [self.emptyViewBottomConstraint setConstant:-self.currentKeyboardHeight];
```

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -1455,34 +1519,21 @@ - (void)keyboardWillHideNotification:(NSNotification *)notification
...
         [self.collectionView.collectionViewLayout invalidateLayout];
     }];
 }
 
 /**
- Centers the empty view taking into account the collection view height and content insets.
+ Adjusts empty view to refresh constraints based on the main view.
  */
-- (void)centerEmptyView
+- (void)adjustEmptyView
```
hassaanelgarem: This function name implies that we're doing more than just refreshing the layout. I'd consider removing it and just calling `[self.view layoutIfNeeded]` directly when needed.

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -602,14 +631,14 @@ - (UIView *)emptyView
     return _emptyView;
 }
 
-- (void)addEmptyViewToView
+/** Checks if the parentViewController is providing a custom empty ViewController to be added, if not, add a provided custom emptyView
```
hassaanelgarem, line 627: Thank you for adding the documentation. Can we also mention that `emptyViewControllerForMediaPickerController` takes precedence over `emptyViewForMediaPickerController` in the public documentation of these two functions in `WPMediaPickerViewController.h` 🙏

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: You're right, makes more sense 😄 ... I have a big problem naming things 😅 (One of the hardest things in the IT area haha) this is one of the reasons that I love the reviews on PR, pointing things that pass unnoticed for who is doing the changes, thanks! I'll rename it to `addEmptyViewToContainer`

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: Let me do some tests with this behavior for hide/show, I don't think we have drawbacks to follow with it.

WPMediaPicker.podspec
diogo-lins: Resolved here https://github.com/wordpress-mobile/MediaPicker-iOS/pull/399/commits/ddf67174b0aeaf8db5716194e27767a2f0613d34 😄

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: Resolved here https://github.com/wordpress-mobile/MediaPicker-iOS/pull/399/commits/e54f6a1a818e1226233fcbca6bb9649e57447b42 😄

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: Resolved here https://github.com/wordpress-mobile/MediaPicker-iOS/commit/e54f6a1a818e1226233fcbca6bb9649e57447b42 😄

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: Resolved here https://github.com/wordpress-mobile/MediaPicker-iOS/commit/e54f6a1a818e1226233fcbca6bb9649e57447b42 😄

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: Resolved here https://github.com/wordpress-mobile/MediaPicker-iOS/commit/e54f6a1a818e1226233fcbca6bb9649e57447b42 😄

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: Documentation updated here https://github.com/wordpress-mobile/MediaPicker-iOS/pull/399/commits/72ba07e4ceeca34ac4093fe4b6b11cdc8e5b8363 😄

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: Worked smoothly 😄, resolved here https://github.com/wordpress-mobile/MediaPicker-iOS/pull/399/commits/eaf7f546a638a29a556d00d97ec15941d1e4492d

diogo-lins:
> The single media asset issue is resolved 🚀
> 
> > we need to update the constraint on the NoResultsViewController for this case, this VC that manipulates the text top constraint, should I create an issue and raise the new PR there after this PR be merged and bump the pod version on the WordPress project?
> 
> There's no need to block this fix until #399 is merged, we should actually aim to test both concurrently.
> 
> Please create a PR on WPiOS with this fix before we merge this PR. For now, the WPiOS PR should point to the `WPMediaPicker` living on the branch `issue/18840-19514-emptyview-height-v2`. Once #399 is merged and the new version is released, we can update the WPiOS PR to point to the released version.

Got it! Let me do this 🚀

Pod/Classes/WPMediaPickerViewController.h
```diff
@@ -222,6 +222,9 @@
  *  @return An empty view to display or `nil` to not display any.
  *
  *  If this method is not implemented, a default UILabel will be displayed.
+ *
+ *  In case of (emptyViewControllerForMediaPickerController) and (emptyViewForMediaPickerController) are both implemented by a parent ViewController,
+ *  the first one has precedence over the second one to be presented to the user.
```
hassaanelgarem: What about a simpler statement like this one.
```suggestion
 *  `emptyViewControllerForMediaPickerController` takes precedence over this method.
```

Pod/Classes/WPMediaPickerViewController.h
```diff
@@ -235,6 +238,9 @@
  *
  *  If this method is not implemented, a default ViewController with a default
  *  UILabel will be displayed.
+ *
+ *  In case of (emptyViewControllerForMediaPickerController) and (emptyViewForMediaPickerController) are both implemented by a parent ViewController,
+ *  the first one has precedence over the second one to be presented to the user.
```
hassaanelgarem:
```suggestion
 *  This method takes precedence over `emptyViewForMediaPickerController`.
```

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -945,7 +923,7 @@ - (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSe
 - (void)toggleEmptyViewFor:(NSInteger)numberOfAssets
 {
     if (numberOfAssets > 0) {
-        [self removeContainerEmptyView];
+        self.emptyViewContainer.hidden = true;
     } else {
         [self addContainerEmptyView];
```
hassaanelgarem: What about moving `[self addContainerEmptyView]` to `viewDidLoad`, and here only hide or show `emptyViewContainer`?

Pod/Classes/WPMediaPickerViewController.m
```diff
@@ -592,28 +592,29 @@ - (void)showCapture {
  */
 - (void)addContainerEmptyView
 {
-    if (self.containerEmptyView != nil && self.containerEmptyView.superview != nil) {
-        [self selectEmptyView];
+    if (self.emptyViewContainer != nil && self.emptyViewContainer.superview != nil) {
+        self.emptyViewContainer.hidden = false;
+        [self addEmptyViewToContainer];
```
hassaanelgarem: I think we can now get rid of these checks since we'll be sure we'll only call this once. Same for the checks in `addEmptyViewControllerToView` and `addEmptyViewToView`. What do you think?

Pod/Classes/WPMediaPickerViewController.m
hassaanelgarem: I'm sorry that I wasn't clear enough. I was suggesting renaming the four existing functions to these four new names. What I was suggesting was:
- `addContainerEmptyView` -> `addEmptyViewContainer`
- `addEmptyViewToContainer` -> `populateEmptyViewContainer`
- `addEmptyViewToView` -> `addEmptyViewToContainer`
- `addEmptyViewControllerToView` -> `addEmptyViewControllerToContainer`

Pod/Classes/WPMediaPickerViewController.h
diogo-lins: Agreed 😄!

Pod/Classes/WPMediaPickerViewController.h
diogo-lins: Agreed 😄!

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: Ohhh, I got it. Let me rename these.

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: I did some tests locally here, and I have some discoveries: If the `parentView` is providing an `emptyViewController` after a search made by the user, it calls the [self.collectionView reloadData]` removing the `self.emptyViewController.view` from subviews. 

So, even calling the `[self.emptyViewContainer setHidden: YES | NO]` shows only the container without subviews. It happens because the `view` comes from an `addChildViewController`, and calling `didMoveToParentViewController`, the view is removed from subviews for some reason, maybe after the `collectionView` reloads the data.

Take a look on the moments:
Screen Shot 2022-11-03 at 15 19 56
Screen Shot 2022-11-03 at 15 20 24

I searched on the codeBase for codes using the `didMoveToParentViewController` specific for the `noResultsViewController`, and it seems all of them remove it from `superview`, instead hide.

It works if we move the `[self populateEmptyViewContainer]` to:
```
- (void)toggleEmptyViewFor:(NSInteger)numberOfAssets
{
    if (numberOfAssets > 0) {
        self.emptyViewContainer.hidden = true;
    } else {
        self.emptyViewContainer.hidden = false;
        [self populateEmptyViewContainer];
    }
}
```
With this approach, we can leave the `[self addEmptyContainerView]` on `viewDidLoad`, but needs to call the populate in the toggle moment. What do you think about that?

I think this was why the older behavior was using `remove/add` the `emptyViewController` at this moment instead `show/hide` logic.

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: I agree to remove for `emptyViewContainer`, it works fine, but for the `addEmptyViewControllerToView` and `addEmptyViewToView` we need it, if you consider my previous comment about the `viewDidLoad`, with this, we avoid adding two views on the container.

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: I did some tests locally here, and I have some discoveries: 
The `parentView (GutenbergViewController)` provides an `emptyView (NoResultsViewController)`. 
When the `WPMediaPickerViewController` starts the lifecycle and advises the `parentView` from a callback `(didUpdateSearchWithAssetCount:)` that should hide the `emptyView`, in turn, `NoResultsViewController` removes its view from the superView at that moment.

Take a look at the moments:
Screen Shot 2022-11-03 at 15 19 56
Screen Shot 2022-11-04 at 16 05 47
Screen Shot 2022-11-04 at 16 06 08
Screen Shot 2022-11-04 at 16 06 49

So, even using hide and show is not enough for our case; Since this `NoResultsViewController` is used in other parts of the App, I don't think it is an option to change the logic in there to don't cause some side effects in other areas.

We can follow this approach, move the `[self populateEmptyViewContainer]` to:
```
- (void)toggleEmptyViewFor:(NSInteger)numberOfAssets
{
    if (numberOfAssets > 0) {
        self.emptyViewContainer.hidden = true;
    } else {
        self.emptyViewContainer.hidden = false;
        [self populateEmptyViewContainer];
    }
}
```
With this, we can leave the `[self addEmptyContainerView]` on `viewDidLoad`, but it needs to call the populate in the toggle moment. What do you think about that?

Pod/Classes/WPMediaPickerViewController.m
diogo-lins: I agree with getting rid of the `emptyViewContainer` creation moment, works fine, but for `addEmptyViewControllerToView` and `addEmptyViewToView` we should keep it, to avoid adding multiple children views, if you consider my other comment about the `viewDidLoad` https://github.com/wordpress-mobile/MediaPicker-iOS/pull/399#discussion_r1014394325

Pod/Classes/WPMediaPickerViewController.m
hassaanelgarem: Thanks for the deep dive @diogo-lins! I think your recommendation makes sense 👌

hassaanelgarem (approved):
Works as described 🚀

PR #409, opened 2023-05-12
Fix crash when displaying an empty view

hassaanelgarem:
## References
- [WPiOS Issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/19965)
- [WPiOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/20666)

## Description
If the media picker delegate implements the `emptyViewController` function but ends up returning nil, this results in a crash. This PR makes sure that the empty view controller is always set, even if the delegate returns nil. 

## Testing Details
Please test using the referenced WPiOS PR.

---

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - WPMediaPicker (1.8.7)
+  - WPMediaPicker (1.8.8-beta.1)
 
 DEPENDENCIES:
   - WPMediaPicker (from `../`)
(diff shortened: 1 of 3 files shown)


momo-ozawa (approved):

staskus:
Unfortunately, it caused [WPMediaPickerViewController crash: NSInternalInconsistencyException: Invalid batch updates detected](https://github.com/wordpress-mobile/WordPress-iOS/issues/21102), we'll need to revert or adjust these changes.

PR #411, opened 2023-07-14
Fix crash when the same media picker instance gets pushed twice

### LINKED ISSUE #20890: NSInvalidArgumentException: <UINavigationController: 0x1370f9400> is pushing the same view controller instance
This corresponds to one of the distinct crashes found in [JETPACK-IOS-31](https://a8c.sentry.io/issues/3140539221/?referrer=github_integration):

```
NSInvalidArgumentException
<UINavigationController: 0x1370f9400> is pushing the same view controller instance (<WPMediaPickerViewController: 0x138021a00>) more than once which is not supported and is most likely an error in the application : com.automattic.jetpack
```

kean:
Fixes [#20890](https://github.com/wordpress-mobile/WordPress-iOS/issues/20890) reported in WP-iOS.

I haven't found any ways to reproduce it, but this defensive code should address the crash. If the same `UIViewController` is pushes to the navigation stack twice, it does crash every time.

The changes can be tested in https://github.com/wordpress-mobile/WordPress-iOS/pull/21105

---

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - WPMediaPicker (1.8.8)
+  - WPMediaPicker (1.8.9-beta.1)
 
 DEPENDENCIES:
   - WPMediaPicker (from `../`)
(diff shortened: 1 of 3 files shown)


guarani:
👋 Please re-request a review when this is ready since from https://github.com/wordpress-mobile/WordPress-iOS/pull/21105#issuecomment-1636621068, it sounds like you're still working on this. Thanks @kean!

PR #414, opened 2023-08-14
Avoid WPMediaPickerViewController reloadItemsAtIndexPaths crash by wrapping the call in try-catch block

staskus:
Fixes #21102

This is an extension of https://github.com/wordpress-mobile/MediaPicker-iOS/pull/412 fix.

There're still some crashes happening in `WPMediaPickerViewController`. Although `updateDataWithRemoved` is alredy called from within try-catch block, the logic inside the completion block happens asynchronously and happens outside the scope of try-catch block`

To test:

I couldn't reproduce the original crash, however, adding a fake `insertItem..` calls in the completion block I was able to reproduce similar crashes and confirm that try-catch block helps to avoid them temporarily. 

The best way to avoid these crashes long term is to adopt UICollectionViewDiffableDataSource which requires refactoring of the current solution.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,12 @@ _None._
 
 _None._
 
+## 1.8.10
+
+### Bug Fixes
+
+- Fix WPMediaPickerViewController crash when reloading collection view items [#414]
+
 ## 1.8.9
 
 ### Bug Fixes
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - WPMediaPicker (1.8.9-beta.1)
+  - WPMediaPicker (1.8.10-beta.1)
 
 DEPENDENCIES:
   - WPMediaPicker (from `../`)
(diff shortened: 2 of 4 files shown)


guarani (approved):
Hey @staskus, I agree with this approach since we can't reproduce this and it's an extension of the same temporary fix we've applied recently. 

Looking at the three crashes, the logs indicate the data source's first section is emptied but no removals in the updates. They also all happen on the call to `reloadItemsAtIndexPaths`. Since the call to this method doesn't involve updates (it's the prior call to `performBatchUpdates` that does), I'm not sure I understand the error message. 

After reading https://stackoverflow.com/questions/38597804/how-to-order-moves-inserts-deletes-and-updates-in-a-uicollectionview-performb, it could be the case that we're reloading "old" index paths which don't correspond to a data source item after the batch updates. 

I tried simulating a few move scenarios using the example library but couldn't make it crash. I think this is fine to merge as-is.

guarani:
> The best way to avoid these crashes long term is to adopt UICollectionViewDiffableDataSource which requires refactoring of the current solution.

That's true. What also came to find is https://github.com/wordpress-mobile/WordPress-iOS/issues/21190 😄

PR #420, opened 2024-01-29
Allow camera use with limited library permission

### LINKED ISSUE #419: Cannot add image from camera using 1.8.3 onwards, when a limited subset of the library is selected.
**Describe the bug**
When attempting to add an image using the camera, the following error is received:

> Cannot upload image – the operation couldn't be completed. (WPMediaPickerErrorDomain error 0.)

**To Reproduce**
This was seen in the Woo app, after updating to 1.8.10.
1.8.1 and 1.8.2 do not have the bug.

Update the Woo app's Podfile to specify any version of WPMediaPicker after 1.8.2

When asked for photo library permission, pick a limited selection of the library. Note that the user may have previously made this selection, and won't necessarily be prompted to give permission again.

1. Launch the app
2. Log in to a store
3. Tap the Products tab
4. Select a product
6. Tap the `Add` button in the images row at the top (scroll the row left if needed)
7. Tap `Add Photos`
8. Tap `Take a photo`
9. Take a photo
10. Tap `Use Photo`
11. Observe that the error `Cannot upload image – the operation couldn't be completed. (WPMediaPickerErrorDomain error 0.)` is shown

For me, this is 100% reproducible.

**Expected behavior**

The error should not be shown, and the image should be uploaded as usual.

**Screenshots**
failed-to-upload-from-camera

**Smartphone (please complete the following information):**
 - Device: iPhone 15 Pro Max
 - OS: iOS 17.3 (public beta)
 - Version: 1.8.2+

**Additional context**

I've tested with 1.8.1, 1.8.2, 1.8.3, 1.8.5, and 1.8.10.
Every version I tried after 1.8.2 has the issue. [1.8.2 has a fix](https://github.com/wordpress-mobile/MediaPicker-iOS/pull/383) for [another related permissions issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/17467)

[Changing `-[WPPHAssetDataSource addAssetWithChangeRequest:completionBlock:]`](https://github.com/wordpress-mobile/MediaPicker-iOS/blob/3b7ff045ec5c051df821b68b8d8bc0af8e69a06c/Pod/Classes/WPPHAssetDataSource.m#L466) to handle both 
```
            case PHAuthorizationStatusAuthorized:
            case PHAuthorizationStatusLimited:
```
using the existing `PHAuthorizationStatusAuthorized` handler, instead of limited being treated the same as `PHAuthorizationStatusDenied`, fixes the issue.

That makes sense, and matches how we handle these two permissions in [`loadDataWithOptions`](https://github.com/wordpress-mobile/MediaPicker-iOS/blob/3b7ff045ec5c051df821b68b8d8bc0af8e69a06c/Pod/Classes/WPPHAssetDataSource.m#L161).

Various warnings can be seen in the console when using the camera:

> Attempted to change to mode Portrait with an unsupported device (BackWideDual). Auto device for both positions unsupported, returning Auto device for same position anyway (BackAuto).
> Attempted to change to mode Portrait with an unsupported device (BackWideDual). Auto device for both positions unsupported, returning Auto device for same position anyway (BackAuto).
> Attempted to change to mode Portrait with an unsupported device (BackAuto). Auto device for both positions unsupported, returning Auto device for same position anyway (BackAuto).
> Attempted to change to mode Portrait with an unsupported device (BackWideDual). Auto device for both positions unsupported, returning Auto device for same position anyway (BackAuto).

However these show both in versions with the bug, and versions where camera images work.

joshheald:
Fixes #419

## Description

Previously, any attempt to add an image from the camera when you had only granted `PHAuthorizationStatusLimited` photo library permission resulted in an error message.

The example app handled that by directing users to go to settings and change their permission… but that’s not very obvious, after all, they’re not trying to add something from their library. Apple does not require library authorization for use of the camera.

This PR changes the `limited` permission to be handled identically to `authorized`, which removes the error message.

## To test:

1. Launch the example app
2. Tap the `Tap here to quick select assets` row
3. When prompted, grant limited access to one or more photos in your library
4. When prompted, grant camera access
5. Use the camera interface to take a photo
6. Tap `Use photo` – note that you do not see an error at this point.
7. Select the photo in the collection view
8. Tap `Done`
9. Observe that it's added to the list

## Screenshots

https://github.com/wordpress-mobile/MediaPicker-iOS/assets/2472348/d8066250-ea2e-410c-9bde-dabfbe96dd88

---

### DIFF
--- a/WPMediaPicker.podspec
+++ b/WPMediaPicker.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WPMediaPicker'
-  s.version       = '1.8.11-beta.1'
+  s.version       = '1.8.10'
 
   s.summary       = 'WPMediaPicker is an iOS controller that allows capture and picking of media assets.'
   s.description   = <<-DESC
--- a/WPMediaPicker.podspec
+++ b/WPMediaPicker.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WPMediaPicker'
-  s.version       = '1.8.10'
+  s.version       = '1.8.11-beta.1'
 
   s.summary       = 'WPMediaPicker is an iOS controller that allows capture and picking of media assets.'
   s.description   = <<-DESC
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ _None._
 
 ### Bug Fixes
 
-_None._
+- Allow camera use when limited photo library permission is granted [#419]
 
 ### Internal Changes
 
--- a/Pod/Classes/WPPHAssetDataSource.m
+++ b/Pod/Classes/WPPHAssetDataSource.m
@@ -463,7 +463,6 @@
                 return;
             }
             case PHAuthorizationStatusDenied:
-            case PHAuthorizationStatusLimited:
             {
                 if (completionBlock) {
                     dispatch_async(dispatch_get_main_queue(), ^{
@@ -481,6 +480,7 @@
                 return;
             }
             case PHAuthorizationStatusAuthorized:
+            case PHAuthorizationStatusLimited:
             {
                 dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INITIATED, 0), ^{
                     [[PHPhotoLibrary sharedPhotoLibrary] performChanges:^{


WPMediaPicker.podspec
```diff
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WPMediaPicker'
-  s.version       = '1.8.10'
+  s.version       = '1.8.11-beta.1'
```
jaclync: From my understanding of the recent streamlined process paaHJt-5zM-p2 (also linked from the iOS handbook P91TBi-2EK-p2), we no longer need to update the pod version during development. We'll just point to the (merge) commit from the client repo, and the apps infra team will ship & update the library plus the client repo.

jaclync (approved):
Was able to see the permission alert in trunk following the testing steps, and it was fixed in this PR! :shipit: 

before | after
-- | --
 | 


For some reason I couldn't reproduce this in WCiOS trunk, the limited photo library modal was shown instead of an error. With the PR https://github.com/woocommerce/woocommerce-ios/pull/11843, the UX remained the same for me which is all good, not blocking image upload.

UX in WCiOS trunk & PR:


https://github.com/wordpress-mobile/MediaPicker-iOS/assets/1945542/d8b564c1-48db-45a3-b8e9-d204ce3d8b4b

WPMediaPicker.podspec
joshheald: Thanks! I didn't know/had forgotten about that, but it's much easier.

PR #424, opened 2025-02-28
Issue/423 prevent index overflows

### LINKED ISSUE #423: Crash: out of range values passed to `[NSMutableIndexSet addIndexesInRange:]` when called from `WPPHAssetDataSource photoLibraryDidChange:`
**Describe the bug**
In Woo iOS, we have a crash which occasionally happens to users with an out of range photo. We're unable to reproduce it, but can see where it might be happening.

NSIndexSets are used to get the index, which are enumerated using `[indexes enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL * _Nonnull stop)]`. Note that `idx` is an unsigned integer.

idx is then passed to - (NSInteger)adjustedIndexForIndex:(NSInteger)index forCount:(NSInteger)count, which expects a signed integer, and manipulates it by subtracting 1 in some cases to reverse the order.

The indexes we're seeing in crash reports look like overflows, e.g. `18446744073709551614` is common in the crash logs, which is equal to `ULONG_MAX - 1`, or `(2^64 - 1) - 1`.

This suggests we're getting instances where idx is greater than count, and not guarding against them before doing our arithmetic.

It seems that we should adjust our `adjustedIndexForIndex` function to use `NSUIntegers` in all cases, or failing that, guard for out of range values. Since they're indexes, they should never be negative anyway.

There is some protection against this issue when we're loading cells – WPMediaPickerViewController.assetForPosition has the following check before calling it:

```
if (itemPosition >= count || itemPosition < 0) {
        return nil;
}
```

However that code path isn't used in this crash; it comes from the libraryDidChange notification from iOS. Again, we're passed NSUIntegers from here.

**To Reproduce**
Unknown. I've tried changing the photo library with the app open at the picker, and approaches changing it with the partial access. I think it probably relies on some strange device state as it's fairly uncommon.

**Expected behavior**
It shouldn't crash.

**Screenshots**
None

**Smartphone (please complete the following information):**
 - Device: various iPhones
 - OS: any iOS
 - Version: any

**Additional context**
This happens even with the latest version of WPMediaPicker.

joshheald:
Fixes #423 

This is to resolve a crash which happened when `-[WPPHAssetDataSource photoLibraryDidChange:]` was called,with an invalid (seeming) `changedIndexes` value.

To test:

I wasn't ever able to actually reproduce the crash, but I believe it happened due to a combination of:
1. Indexes which were greater than the count, leading to:
2. Overflows to a very high UInt when doing the arithmetic in `-[WPPHAssetDataSource adjustedIndexForIndex:forCount:]`
3. Those not being properly checked before they were passed to `-[NSMutableIndexSet addIndex:]`, which requires that the UInts it's passed be between 0 and NSIntegerMax -1.

I've made a few tweaks to avoid that:
1. Throw an exception if we try to adjust an index which is greater than the count. It will lead to a negative number, which makes no sense for an index
2. Don't even attempt to adjust the index in `-[WPPHAssetDataSource adjustedIndexesForIndexSet:forCount:]` if it's greater than the count.
3. Use Apple's record of the count of assets before the changeset, not our `assetCount` which may be out of date.

I've tested it by adding to the indexes of any changes in `photoLibraryDidChange`, before passing them to the adjust functions, and checking that it doesn't crash. I tested in the Woo app.

Testing steps:
1. Launch the app using the new version of MediaPicker
2. Open a product, and add an image to it
3. Go to the camera and take some photos
4. Go to Photos and edit some of the recent photos, and delete some too
5. Go back to woo and check it doesn't crash

---

### DIFF
--- a/WPMediaPicker.podspec
+++ b/WPMediaPicker.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WPMediaPicker'
-  s.version       = '1.8.11'
+  s.version       = '1.8.12'
 
   s.summary       = 'WPMediaPicker is an iOS controller that allows capture and picking of media assets.'
   s.description   = <<-DESC
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,12 @@ _None._
 
 _None._
 
+## 1.8.12
+
+### Bug Fixes
+
+- Prevent crash when unexpected indexes received in a PHAsset changeset [#423]
+
 ## 1.8.11
 
 ### Bug Fixes
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - WPMediaPicker (1.8.11)
+  - WPMediaPicker (1.8.12)
 
 DEPENDENCIES:
   - WPMediaPicker (from `../`)
@@ -9,7 +9,7 @@ EXTERNAL SOURCES:
     :path: "../"
 
 SPEC CHECKSUMS:
-  WPMediaPicker: 79548964e9c7e131ccfa72270bdda3aec8fa56b5
+  WPMediaPicker: e9eaa804e1b0288d7969776608053ae0ea2941f2
 
 PODFILE CHECKSUM: 31590cb12765a73c9da27d6ea5b8b127c095d71d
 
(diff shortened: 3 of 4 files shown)


WPMediaPicker.podspec
AliSoftware, line 1: ⚠️  You need to run `bundle exec pod install` in the `Example` project to also update [its `Podfile.lock`](https://github.com/wordpress-mobile/MediaPicker-iOS/blob/trunk/Example/Podfile.lock) so it accounts for that version bump

itsmeichigo (approved):
I tested Woo app running this change on an iPhone 16 Pro iOS 18.3 and confirmed that the app didn't crash after changing photos in the library.

